PR #623, opened 2026-03-02
Updated ng-wrapper scripts and packaging

gnimmelf:
NOTE! Used modal window component to check that this works, so it contains a few more props for hiding cancel and close buttons.

Main changes is to `npm scripts` - mostly a cleanup and som higher level scripts running multiple others.
Secondary changes is to ng-wrapper post gen fixup scripts before ng-wrapper is actually `npm install`ed and `npm build`t.


tibnor (approved):
Thanks for the PR!

I removed the patches, these can be separate PRs. We have an ongoing discussion on styling in #579. I like the idea of having some run scripts in the root package.json, for beginners of the project to get started.

tibnor:
The path error is fixed in https://github.com/lit/lit/pull/5313, waiting on approval

talpitoo:
@gnimmelf  after https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/1166 landed, you might need to rebase this having those new code&comment conventions in mind.

talpitoo:
Related: #838 (tree-shaking discussion, `ng-wrapper-treeshaking` branch) and #1128 (wrappers now use oicl-lit packages instead of tarballs), which changed the same scripts.

PR #663, opened 2026-03-13
feat: upgrade readout component 

anastasiyaBoyko:
## Summary by CodeRabbit

* **New Features**
  * New readout components for navigation with horizontal/vertical layouts, multiple sizes, and enhanced/regular/stack variants.
  * Segmentized input/advice/value elements with optional icons, button mode, and active/amplified states.
  * Fixed-width value handling, hinted-zero padding, degree suffix support, and improved text truncation for fixed-length labels/units.
  * Source display with optional picker/flyout and selectable source items.
  * Expanded typography mixins and comprehensive styling for consistent readout layouts and Storybook showcases.


talpitoo (commented):
@anastasiyaBoyko i haven't worked outside of 'Instruments' yet (apart from circular progress), so take these review comments as 'my learning curve questions', not necessarily requested changes.

1. I can see that we have a very similar component https://openbridge-jip-storybook.web.app/?path=/docs/instruments-instrument-field--docs, shouldn't this be an extension/update on that one (referenced e.g. from https://openbridge-jip-storybook.web.app/?path=/docs/instruments-azimuth-thruster-labeled--docs)?

2. Playing around with `obc-component-size-*` at https://openbridge-storybook--663-feat-upgrade-readout-compone-aiy09snn.web.app/?path=/story/instruments-readout--playground&globals=componentSize:obc-component-size-medium selecting 'large' from the dropdown hides the advice triangle. Selecting 'XL' scrambles the layout as if some parts of it are either hidden or outside of the viewport.

3. Which is the correct Figma link to compare to?

4. When i click the 'Source picker' and click outside, it won't 'blur' although that might be related to the dropdown component.

5. I can see variations like 'Advice + Degree' but no `°` symbol. Maybe it looks as intended, i just don't have the full context for review yet.

6. Question: stories with 'flyout' have something like `SRC >`. Is that all or should 'something fly out' e.g. on hover as the name suggests?

7. Haven't looked into them, but perhaps scan through coderabbit suggestions? :rabbit: 

More to come...

Anyway, it looks good after a first look and i am doing this review as @tibnor asked but it is more a 'learning curve for myself' than an actual review for now. Thanks.

anastasiyaBoyko:
@talpitoo thanks!
1. On this PR I’m keeping the existing obc-instrument-field because it’s effectively the previous/legacy version of this pattern, and while the PR is still in draft I’m using it as a baseline for comparisons and spot-checking during implementation (layout/spacing/formatting and general behavior).
I did start from obc-instrument-field as the base and upgraded the approach, but during the work it became clear we shouldn’t grow instrument-field further: we decided to split the “input” and “advice” parts into dedicated components and to upgrade the overall composition as readout. That keeps concerns separated, avoids forcing a breaking redesign of the legacy component in the same PR, and makes the new readout system reusable across instruments.

ps checking more..

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
```diff
@@ -0,0 +1,616 @@
...
+
+  @property({type: String}) inputDescription = '';
+
+  @property({type: String}) inputValueLength = '';
+
+  @property({type: Boolean}) inputHasHintedZeros = false;
+
+  @property({type: Boolean}) inputHasDegree = false;
+
+  @state() private sourcePickerContentVisible = false;
```
talpitoo, line 251: In readout, the open state is owned entirely by `obc-readout`:

- the SRC trigger only toggles that boolean through `onTogglePicker`
- the menu is rendered only when that boolean is true
- it closes only after an item-click

There is no outside-click, blur, focusout, or popover-based dismissal in readout.

Examples where it is working as intended:
- https://openbridge-storybook.web.app/?path=/story/ui-components-buttons-menu-button--regular
- https://openbridge-storybook.web.app/?path=/docs/ui-components-buttons-split-button--docs

@anastasiyaBoyko this is the same as [my other comment under 4.](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/663#pullrequestreview-4146590904), just looked into it in more detail.

Suggested fix:
- keep `sourcePickerContentVisible` as the open state
- add a window-level `pointerdown` handler when opening
- close if the click target is outside the readout
- remove the listener on close/disconnect

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
```diff
@@ -0,0 +1,616 @@
...
+  supportsReadoutSourcePicker,
+} from './readout-source.js';
+import {
+  ReadoutInputState,
+  ReadoutInputType,
+  ReadoutInputVariant,
+} from '../readout-input/readout-input.js';
+import '../readout-input/readout-input.js';
+
+export enum ReadoutType {
```
talpitoo: The readout API overloads ‘enhanced’ in a way that does not match the existing navigation-instrument [Priority](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/blob/develop/packages/openbridge-webcomponents/src/navigation-instruments/types.ts#L13) / [priorityElements](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/blob/develop/packages/openbridge-webcomponents/src/navigation-instruments/compass-flat/compass-flat.ts#L82) model. In instruments, `Priority.enhanced` is a scoped color-emphasis concept; in readout, `enhanced` currently means layout style, type preset, and visual state/color. If readout is intended for reuse inside priority-driven instruments, I’d recommend separating those concepts in the API: keep layout/presentation style separate, and introduce a [priority](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/blob/develop/packages/openbridge-webcomponents/src/navigation-instruments/azimuth-thruster-labeled/azimuth-thruster-labeled.ts#L65)-based emphasis API for color/state so it composes cleanly with the rest of the instrument system.

I feel we might have to discuss this with @tibnor and **@Jon_Fauske.**

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
anastasiyaBoyko: fixed

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
anastasiyaBoyko: Good point. In the instrument system, Priority.enhanced is a scoped color-emphasis concept (optionally via priorityElements). In obc-readout, enhanced currently refers to the Figma “Enhanced” variant (a presentation/layout preset), so it also affects spacing/typography—not color.
To avoid conflating these concepts, I can keep the readout variant API (regular/enhanced/stack) and introduce a separate priority/emphasis API for color/state (e.g. a priority/emphasis prop, with optional scoped mapping later if needed). This will let obc-readout compose cleanly inside priority-driven instruments.
Also worth noting: today regular/enhanced/stack in obc-readout behaves primarily as a composition/layout preset (structure and sizing). It does not intentionally map to the instrument Priority.enhanced color-emphasis model, and most color/state styling is currently driven by segment state/tokens rather than the readout type itself
Знімок екрана 2026-04-22 о 13 37 30

anastasiyaBoyko:
@coderabbitai please review the latest changes.

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
talpitoo: @anastasiyaBoyko i can verify, it works with mouse but still not accessible via keyboard. I don't want to sound nitpicking and in general i am not familiar with OpenBridge's status quo regarding WCAG accessibility but in the 'ideal world' a similar pattern would work e.g. https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

- tab to the component
- press space/ENTER to open it
- use arrow keys, or at least tab through the options

cc. @tibnor

anastasiyaBoyko: sure, let me check

talpitoo: @anastasiyaBoyko just started a thread about this, i think it is unrelated to the 'readout' itself so i guess that my comment is out of scope of this review.

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
talpitoo: @anastasiyaBoyko a video is worth a thousand words (multiplied by 24 FPS): a demo of how priority works in the instruments https://www.loom.com/share/ffb1fe979aa4414dbd3a905a50b355d2 (a video version of my initial comment)

anastasiyaBoyko: @talpitoo great, thanks!

talpitoo: ...and copy-pasting Jon's comment [from Slack](https://openstudio-global.slack.com/archives/C09NGQV83BJ/p1776855490945139) so that everything is in one place:

Yep, that needs some cleanup i think... not so easy to find a good system. It would be great to align it with the instruments!
The readouts can have different size (i think we should detach size from enhancement, so a 24px font can be bort regular and enhanced).
Priority makes sense to align with the instrument pris; Regular, enhanced, low integrity etc.. Input is a bit difficult, becaus any priority can have input[12:58 PM]so input can be its own enum maysbe because inputs have differen formats

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
talpitoo: @anastasiyaBoyko keyboard a11y was fixed in https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/811. If you rebase this PR on top of the latest `develop` i guess the issue will be automatically resolved here as well (FYI).

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
anastasiyaBoyko: @talpitoo 
@tibnor FYI

Added fixes based on the discussion about API, separating layout and priority concepts.

### What was changed

#### 1. Separation of concerns in API

The API is now clearly split into independent axes:

- `variant` (regular | enhanced | stack)
  → **layout/composition only**
  → controls structure, spacing, arrangement

- `priority` + `priorityElements`
  → **color/emphasis only**
  → aligned with instrument Priority model

- segment-specific props:
  - `inputPriority`
  - `advicePriority`
  → allow scoped emphasis per segment

---

#### 2. Size decoupled from "enhanced"

Typography is no longer implicitly tied to `enhanced`.

- same size can exist in:
  - regular
  - enhanced
  - stack

This follows the Jon feedback:
> size should not be coupled to enhancement

---

#### 3. Segment-driven behavior clarified

Readout is now a **composition of segments**, each responsible for its own behavior:

- value
- input
- advice
- source

Each segment:
- has its own format (e.g. button, description, range)
- has its own state (display / input / temporary)
- can have its own priority (color emphasis)

No implicit coupling from the container.

---

#### 4. Source API improvements

- `source-flyout-click` → unchanged (external trigger)
- `source-change` → added for picker selection (bubbles + composed)
- `sourceDeltaValue` → explicit prop for delta display

This makes source behavior consistent and externally controllable.

---

#### 5. Layout consistency

- Regular / Enhanced / Stack now behave as **layout presets only**
- Segment layout behavior (e.g. hugContent, alignment) is explicit and testable
- Storybook updated to reflect segment-level vs container-level behavior

---

### Result

- API is now aligned with the instrument system
- No more mixing of layout and priority concepts
- Readout is composable inside priority-driven instruments
- Segment behavior is predictable and isolated
- Size, layout, and emphasis are independent axes

anastasiyaBoyko:
@coderabbitai re-review

talpitoo:
flickering

@anastasiyaBoyko the animation on hover is just for the sake of Storybook? Maybe disable it because it is confusing whether it is part of the component CSS or Storybook story CSS only, also it causes [semi-transparency](https://openbridge-storybook--663-feat-upgrade-readout-compone-aiy09snn.web.app/?path=/story/instruments-readout--horizontal-cases-grid) on the dropdown and flickering (see the screenshot).

talpitoo:
Screenshot from 2026-04-28 16-16-40

@anastasiyaBoyko we will have to reuse this component inside e.g. the Compass. How to reproduce the bottom-right readout (123 ROT °/min)? Is that a valid configuration or do we have to ping **@Jon_Fauske**?

anastasiyaBoyko:
> flickering
> @anastasiyaBoyko the animation on hover is just for the sake of Storybook? Maybe disable it because it is confusing whether it is part of the component CSS or Storybook story CSS only, also it causes [semi-transparency](https://openbridge-storybook--663-feat-upgrade-readout-compone-aiy09snn.web.app/?path=/story/instruments-readout--horizontal-cases-grid) on the dropdown and flickering (see the screenshot).

@talpitoo yes, agree, it was only for storybook, removed and pushed already

anastasiyaBoyko:
> Screenshot from 2026-04-28 16-16-40
> @anastasiyaBoyko we will have to reuse this component inside e.g. the Compass. How to reproduce the bottom-right readout (123 ROT °/min)? Is that a valid configuration or do we have to ping **@Jon_Fauske**?

@talpitoo sure you can, just use stack vertical - priority regular for value 
image

packages/openbridge-webcomponents/src/navigation-instruments/readout-input/readout-input.ts
```diff
@@ -0,0 +1,631 @@
...
+
+  @property({type: String}) size: ReadoutInputSize = ReadoutInputSize.small;
+
+  @property({type: String}) format?: ReadoutInputFormat;
+
+  @property({type: String}) mode?: ReadoutInputMode;
+
+  @property({type: String}) priority?: Priority;
+
+  @property({type: Boolean}) hugContent = false;
```
talpitoo: @anastasiyaBoyko at other place it has `reflect: true` but not here but i think the layout depends on it (correct me if i am wrong).

If CSS doesn't use `:host([hugContent])` for input, either drop `reflect` on advice; OR otherwise add it on input.

talpitoo:
@anastasiyaBoyko there are many properties with the `show*` prefix vs. others with `has*` prefix, ignore if minor.

talpitoo:
@anastasiyaBoyko the [a11y PR](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/811) was merged to handle keyboard navigation (e.g. closing the menu on `Esc`) https://openbridge-storybook--663-feat-upgrade-readout-compone-aiy09snn.web.app/?path=/story/ui-components-buttons-menu-button--regular Have you rebased this on top of the latest `develop`? If so any idea why the same component can't be closed with `Esc` in the readout?

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
talpitoo: @tibnor @anastasiyaBoyko it aligns e.g. with the Compass' `priority` and `priorityElements` API but still the wording remains confusing.

E.g. a future `obc-readout-compass` (or compass slotting a readout into its frame) would write something like:

```
<obc-readout
  .priority=${this.priorityFor(CompassPriorityElement.hdg)}
  .variant=${ReadoutVariant.enhanced}
  .inputFormat=${ReadoutInputFormat.enhanced}
></obc-readout>
```

...where all three uses of the word enhanced are independent.

## Recommended cleanup (before another component depends on this API)

Reserve `enhanced` for the priority/color axis only. Rename the layout and format axes:

Today                            | Suggested
---------------------------------|--------------------------------------------------------------------------------------
ReadoutVariant.enhanced          | ReadoutVariant.expanded (or `detailed` / `large`)
ReadoutInputFormat.enhanced      | ReadoutInputFormat.large (or `prominent`)
ReadoutInputState.enhanced       | Drop in favor of priority: Priority (state becomes purely interaction: enabled / input / input-temporary)
ReadoutAdviceState.enhanced      | Same — derive from priority + the existing active / amplified states
priority-enhanced CSS hook       | Keep — this one's accurate

After that, the compass-side template reads exactly as intended:

```
<obc-readout
  .priority=${this.priorityFor(CompassPriorityElement.hdg)}
  .variant=${ReadoutVariant.expanded}
  .inputFormat=${ReadoutInputFormat.large}
></obc-readout>
```

anastasiyaBoyko:
> there are many properties with the `show*` prefix vs. others with `has*` prefix, ignore if minor.

@talpitoo 
This is minor and ok to leave as-is because the prefixes communicate different semantics:
has* flags whether a segment/feature exists in the readout (e.g. hasInput, hasAdvice, hasSrcPicker). They control rendering and layout presence.
show* flags an output/formatting behavior (e.g. showZeroPadding) or is used for derived “should we show X” logic (like internal show* getters for dividers).

tibnor:
What a master piece! This is a huge component! It looks very good!

Since this is a large component I have of course one comments :-) I think formatting of the numbers should be the same for all numbers. So make a global "hasDegree" property which should make the circle for advice, input and value. Also the same for fraction digits. (But not for zero padding and max digits, these can work on the value only.

Else everything looks fine for me

tibnor (approved):

anastasiyaBoyko:
> @anastasiyaBoyko the [a11y PR](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/811) was merged to handle keyboard navigation (e.g. closing the menu on `Esc`) https://openbridge-storybook--663-feat-upgrade-readout-compone-aiy09snn.web.app/?path=/story/ui-components-buttons-menu-button--regular Have you rebased this on top of the latest `develop`? If so any idea why the same component can't be closed with `Esc` in the readout?

@talpitoo fixed, 
obc-context-menu-input fires a close event on esc, but obc-readout wasn’t listening to that event, so its sourcePickerContentVisible state never changed and the menu stayed open

anastasiyaBoyko:
> What a master piece! This is a huge component! It looks very good!
> 
> Since this is a large component I have of course one comments :-) I think formatting of the numbers should be the same for all numbers. So make a global "hasDegree" property which should make the circle for advice, input and value. Also the same for fraction digits. (But not for zero padding and max digits, these can work on the value only.
> 
> Else everything looks fine for me

@tibnor thank you!, fixed:

- Added a single global hasDegree on obc-readout that applies to value + input + advice
- Made fractionDigits shared across value + input + advice (while keeping maxDigits / showZeroPadding value-only)
- Updated obc-readout-input so numbers are formatted with fractionDigits, while strings remain unchanged for backwards compatibility
- Removed the deprecated per-segment degree props (valueHasDegree, inputHasDegree, adviceHasDegree) so hasDegree is the single source of truth

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
```diff
@@ -0,0 +1,1178 @@
...
+@customElement('obc-readout')
+export class ObcReadout extends LitElement {
+  @property({type: String}) variant: ReadoutVariant = ReadoutVariant.regular;
+
+  /**
+   * @deprecated Use `readoutInputStyle` for input emphasis instead of per-element priority selection.
+   */
+  @property({type: String}) priority?: Priority;
+
+  @property({type: String, attribute: 'value-priority'})
```
tibnor: dont use attribute

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.ts
```diff
@@ -0,0 +1,1178 @@
...
+ * @slot unit - Replaces the unit content.
+ * @slot source - Replaces the source row content.
+ * @slot src-picker-content - Provides the source picker context menu content.
+ */
+@customElement('obc-readout')
+export class ObcReadout extends LitElement {
+  @property({type: String}) variant: ReadoutVariant = ReadoutVariant.regular;
+
+  /**
+   * @deprecated Use `readoutInputStyle` for input emphasis instead of per-element priority selection.
```
tibnor: Remove all deprecated apis

anastasiyaBoyko:
@tibnor thanks for review , here is short update:
1. Degree symbol shows up consistently now across all enhanced cases — was a bug in how value text composition dropped the suffix when width templates were active
2. Flip-flop reworked — both input and value always visible, focus swaps based on whether they match. Value not equal setpoint → input large and bold, value small. Value equals setpoint → swap, value becomes large and bold, input small. Added a minimum reserved width so the readout doesn't reflow on focus swap or setpoint reached, same applies to pop-up
3. valueHasFixedLength now reserves a minimum width based on valueLength. Toggling the prop on and off doesn't change layout as long as value fits the template. If value is longer than valueLength template, falls back to natural width so user never sees a misleading display
4. Spacing fixed — was a duplicated width reservation when both valueHasFixedLength and valueHasHintedZeros were true. Template width and hint zero width were both being applied. Now template width clears when hint zeros are present
5. Horizontal padding stays constant when toggling input — input segment width is now reserved at the container level so toggling doesn't trigger padding recalculation
6. Deprecated APIs removed — priorityElements and ReadoutPriorityElement gone, priority renamed to valuePriority, inputValue renamed to setpointValue. Cleaned up unused getters and props across readout, readout-input, azimuth-thruster-labeled, speed-gauge, speed-arrows
7. Vue demo readouts with input migrated to pop-up interaction — Propulsion, PropulsionFerry, Azimuth, InstrumentDemo, VesselMotion, OwnShipData, OwnShipDataCard
8. Vue demo alignment — deferred, want to confirm against updated Figma before committing the layout change. Identified the hypothesis but the as-should-be reference is just a Slack screenshot
9 labelOnly supported, and meta zone conditionally renders label and unit containers so empty meta produces nothing. Works for the css-grid one-column-per-kind use case
Extra stuff done alongside the feedback:
- defaultArgs.hug flipped to true so Playground opens in the realistic compact mode
- Reworked shouldHugNestedSegments so segment stretching matches Figma per variant — Enhanced always stretches nested segments, Regular and Stack follow the outer hug prop
- Added hug value to section and case titles so it's clear at a glance which mode each case demonstrates
- Delta source icon uses neutral color now, matching the rest of neutral source indicators - based on figma

ulrik-jo:
This looks very good! 
A small note though: I see that there are quite some examples/snapshots where the unit is defined as the text "DEG". Since you have added "hasDegree", perhaps you should update the snapshots accordingly?

anastasiyaBoyko:
> This looks very good! A small note though: I see that there are quite some examples/snapshots where the unit is defined as the text "DEG". Since you have added "hasDegree", perhaps you should update the snapshots accordingly?

@ulrik-jo 

Thanks for noticing — the two are actually independent and both correct to keep:
hasDegree controls whether the degree glyph ° is appended to the numeric value itself (e.g. 123 renders as 123°).
unit is the label shown in the meta zone alongside the value— it comes from Figma label spec.
instruments often use both together: a heading readout shows 123° (numeric value with glyph via hasDegree) plus a "HDG DEG" label (heading, degrees) in the meta zone. Other combinations are also valid — 25 with °C unit label and no hasDegree, or 123° with no unit label at all.
The snapshots that use "DEG" as the unit reflect the Figma label convention; they're testing a different prop from hasDegree.

ulrik-jo (approved):

PR #731, opened 2026-04-07
feat: generate screenshots

tibnor:


talpitoo:
Merged `develop` in — 12 conflicts, four months of drift. Answering the obvious question first, then the two things that need your call.

**It does not clash with how snapshots run today.** The export is opt-in behind `VITE_STORYBOOK_TAKE_SCREENSHOT`, so `npm run test-storybook` and CI take the unchanged `visAnnotations` + `vis.setup()` path.

What did need care: `vitest.setup.ts` grew two flake guards on develop that your `if/else` would have bypassed in screenshot mode. I split them by intent —

- the retry error-clearing `beforeEach` stays inside the snapshot branch, since it exists for the vis matcher;
- the font-loading `beforeEach` is left ungated, because exported screenshots want the same deterministic `measureText()` as snapshots do.

`vitest.config.ts` merged on its own: develop's `retry: 3` sits next to your `deviceScaleFactor`.

**1. The size tags are split across two spellings.** develop already had `@ignition-base-width: 200px` (with a colon), and this branch renames some of them to `@snapshot-base-width 200px` (no colon). After the merge, three components still carry the old spelling and `generate-element-meta.ts` only matches the new one, so they silently get `widthPx: null`:

- `rot-linear.ts`, `heading.ts`, `bus.ts`

`@ignition-center` / `@ignition-center-horizontal` are untouched by the script either way. Either rename those three, or teach the regex both spellings — I did not pick for you.

**2. `element-meta.json` is stale and I could not regenerate it.** Running `generate:element-meta` without `screenshots/` present nulls out `screenshotPath`, `widthPx` and `heightPx` for all 257 entries, which is worse than the committed file, so I left the April version in place. It needs a `npm run screenshots` run first.

Also reformatted `.storybook/screenshot.ts` — it was failing `format:check`, which now covers it.

`tsc`, `lint:slots`, `lint:eslint`, `lint:agents`, `format:check` and the `rules` tests pass.

PR #785, opened 2026-04-16
feat: upgrade/create indicators

anastasiyaBoyko:
## Release Notes

* **New Features**
  * Enhanced compass indicator component with new type options (Regular and Labeled) and direction modes (Heading, Course, North) for greater flexibility.

* **Documentation**
  * Updated Storybook examples for compass indicator with expanded story variants showcasing all configuration combinations.


tibnor:
Great work!

Some minor issues:
- https://openbridge-storybook--785-feat-upgrade-create-indicato-50wq2n3n.web.app/?path=/story/indicators-wind-indicator--north-up&args=angle:180;speed:9
Wind direction is typically wind from angle. So angle=0 should point down. So the angle is 180 degree of. I think the name of the prop should be change to windFromAngle
- https://openbridge-storybook--785-feat-upgrade-create-indicato-50wq2n3n.web.app/?path=/story/indicators-wind-indicator--north-up&args=speed:9;northUp:!false
In "northUp=false" the compass direction should be heading or course. I think this should have an rotationAngle as property and this is 0 for northUp=true. So if you have wind from east (windFromAngle=90) and rotationAngle=180 (travels south). The arrow should point right.
- I think the angle should be active in this case: https://openbridge-storybook--785-feat-upgrade-create-indicato-50wq2n3n.web.app/?path=/story/indicators-heading-indicator--xtd&args=angle:142;xtd:0.82
I think the rotation origin should be on the top of the arrow

ulrik-jo (approved):

PR #834, opened 2026-04-27
fix(setpoint)!: unify deadband boundary and extract config mixins

talpitoo:
:warning: please do not merge it yet. I was trying to make the radial instruments more DRY by introducing similar concepts that we've done with setpoint-mixin/setpoint-bundle, but maybe it is low priority for now. Let's wait until some more important PRs get merged first, then i'll rebase it and see if still relevant.

Unify all setpoint / zero-snap math behind the canonical helpers in `svghelpers/setpoint.ts` and extract two new Lit mixins to eliminate copy-paste across the radial instruments.

BREAKING CHANGE: `atSetpoint()` in `thruster.ts` and `instrument-linear.ts` now uses an inclusive deadband (`<=`) instead of the previous strict `<`, matching every other instrument and the canonical `computeAtSetpoint()`. A value sitting exactly on the deadband boundary now reports as "at setpoint". Visual impact is limited to the single-pixel boundary case; all 305 storybook snapshots are unchanged.

Setpoint math (single source of truth)
- thruster.ts, instrument-linear.ts: `atSetpoint()` is now a thin adapter over `computeAtSetpoint()`; the local `<` boundary is gone.
- setpoint.ts: add `isAtZero(value, deadband)` companion helper with strict `<` semantics (deadband-exclusive), plus NaN / non-finite guards. Use it in `deriveRadialSetpointConfig`.
- external-scale.ts (2 sites), thruster.ts (3 sites), instrument-field.ts (1 site): replace inline `Math.abs(...) < deadband` with `isAtZero` / `computeAtSetpoint`.
- Sentinel: `grep -rIn 'Math.abs.*setpoint' src` outside `svghelpers/setpoint.ts` now returns zero matches.

New mixins (zero-behavior-change refactor)
- VisualConfigMixin (`svghelpers/visual-config-mixin.ts`): adds `state`, `priority`, `tickmarkStyle`, `showLabels`, `tickmarksInside`.
- TickmarkIntervalMixin (`svghelpers/tickmark-interval-mixin.ts`): adds `primary/secondary/tertiaryTickmarkInterval` with per-component default cadence via `{defaultPrimary, defaultSecondary, defaultTertiary}` options.
- Migrated: `instrument-radial`, `gauge-radial`, `rudder`, `azimuth-thruster`, `rot-sector` (TickmarkIntervalMixin only). Mixins compose: `TickmarkIntervalMixin(VisualConfigMixin(SetpointMixin(LitElement)), {...})`.

Docs
- `.github/instructions/setpoint.instructions.md`: document `isAtZero`, the `<` vs `<=` distinction, the single-source-of-truth rule, and the three companion mixins.

Verification
- `npm run typecheck`: clean
- `npm run lint:eslint`: clean (4 pre-existing warnings in generated locales)
- 305 / 305 storybook snapshot tests pass across `src/building-blocks` and `src/navigation-instruments`.

### DIFF
--- a/.github/instructions/setpoint.instructions.md
+++ b/.github/instructions/setpoint.instructions.md
@@ -328,6 +328,73 @@ computeAtSetpoint({
 
 ---
 
+## `isAtZero()` — Zero-Snap Helper
+
+Companion to `computeAtSetpoint`, used to decide whether a setpoint that is
+*near zero* should snap to the visual "at zero" state (typically rendered
+in muted colors). Defined in `setpoint.ts`:
+
+```ts
+isAtZero(value: number | undefined, deadband: number): boolean
+```
+
+**Rules:**
+
+- Returns `false` if `value` is `undefined` or `NaN`.
+- Returns `false` if `deadband` is not finite.
+- Otherwise returns `Math.abs(value) < deadband` — note **strict less-than**.
+
+### `<` vs `<=` distinction
+
+| Helper               | Comparison | Rationale                                                                                |
+| -------------------- | ---------- | ---------------------------------------------------------------------------------------- |
+| `computeAtSetpoint`  | `<=`       | Inclusive — the setpoint marker should "lock" at the boundary as soon as it is reached. |
+| `isAtZero`           | `<`        | Exclusive — a value exactly equal to the deadband is still considered nonzero.           |
+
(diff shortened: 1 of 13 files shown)


talpitoo:
Closing temporarily to avoid confusion, might reopen it in the future.

PR #856, opened 2026-05-04
fix: require type array in property

tibnor:
## Summary by CodeRabbit

* **Chores**
  * Added a new ESLint rule to enforce consistent array property declarations across web components.

* **Refactor**
  * Replaced many inline object array types with exported named types for component APIs.
  * Updated array-backed component properties to explicitly declare array metadata, improving type consistency and runtime metadata handling.


talpitoo (approved):
I have pushed a minor fix, other than that LGTM.

ulrik-jo (approved):

PR #864, opened 2026-05-06
feat: improve wind instrument responsiveness

CarelessCourage:
## Context
I'm a consultant at [ECC](https://www.ecc.no/) and a senior frontend engineer from [Olavstoppen](https://www.olavstoppen.no/). We use the open bridge library for one of our apps and I am putting this PR up for suggestion in response to user feedback on our end and to contribute back to the library. 

## Problem
Screenshot 2026-04-30 at 09 02 53

One of our users mentioned that the wind arrow in the wind instrument feels a bit small. We display the wind instrument at medium to small sizes sometimes so improving the legibility of the wind direction at smaller sizes would help us. It would also generally help make the instrument more scalable. 

## Solution
The designers already had a concept ready for dealing with this instrument at various sizes. This PR implements that concept. The wind instrument now has 3 variants. Small, medium, and large. You can set the variant as a prop on the component or you can leave it at auto which automatically switches between the 3 as you change the size of the instrument. 

[Link to the prior Slack discussion and design reference](https://openbridgegroup.slack.com/archives/C06LXTCR269/p1777540594904169?thread_ts=1777538316.743459&cid=C06LXTCR269)

Screenshot 2026-05-07 at 08 32 48
Screenshot 2026-05-06 at 10 47 16
Screenshot 2026-05-06 at 10 48 02

## What Changed
- Wind instrument now have a prop called variant that lets you switch between these varians
- By default its set to auto which uses ResizeObserver and host dimensions to switch variant depending on size
- Visual behavior differs by variant to improve readability at small sizes
- Arrow positioning/sizing is now CSS-driven with container-query-friendly units for responsive behavior
- Storybook stories were updated so small sizes can be tested properly with the width control

## Notes
- I originally made a different solution for this issue but decided to start over to align with design. Heres the[ original PR. ](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/847)

## Summary by CodeRabbit

* **New Features**
  * Wind instrument supports four responsive variants (auto, small, medium, large) with configurable size thresholds and auto-adjust behavior.
  * Storybook now exposes variant and size controls and adds Large/Medium/Small/Auto example stories.

* **Style**
  * Component layout updated to render as a full-width/height block for improved sizing and responsiveness.


talpitoo (changes requested):
Thank you, @CarelessCourage, for the suggestion. Even though this is a draft, I’m requesting “blank” changes to avoid an accidental merge until we get a design/code review. Thanks.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
```diff
@@ -1,14 +1,33 @@
...
+  medium = 'medium',
+  large = 'large',
+}
+
+type ResolvedWindVariant =
+  | WindVariant.small
+  | WindVariant.medium
+  | WindVariant.large;
+
+const WIND_SMALL_MAX_PX = 96;
```
tibnor: Can we make WIND_SMALL_MAX_PX and WIND_MEDIUM_MAX_PX as properties instead? With the default values as set here?

tibnor (commented):
Thank you! For the PR! It is highly appreciated! Could you go through the github copilot suggestions and either resolve them or click "resolved" if it should be ignored? If you are unsure about them just leave a comment.

Again, thanks for your contribution

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
```diff
@@ -17,34 +36,86 @@ export class ObcWind extends LitElement {
...
+      const entry = entries[0];
+      if (!entry) return;
+      const {width, height} = entry.contentRect;
+      const next = resolveAutoWindVariant(Math.min(width, height));
+      if (next !== this._autoVariant) {
+        this._autoVariant = next;
+      }
+    });
+    this._resizeObserver.observe(this);
+  }
```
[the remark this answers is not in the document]
CarelessCourage: Addressed. I added a guard so auto-observation is skipped when ResizeObserver is not available, which avoids errors in SSR/test/older browser contexts.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
```diff
@@ -17,34 +36,86 @@ export class ObcWind extends LitElement {
...
+
```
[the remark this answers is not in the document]
CarelessCourage: Addressed. The observer now runs only when variant is auto. When variant changes away from auto, it is disconnected to avoid unnecessary updates.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
```diff
@@ -17,34 +36,86 @@ export class ObcWind extends LitElement {
...
+  override disconnectedCallback(): void {
+    super.disconnectedCallback();
+    this._resizeObserver?.disconnect();
+    this._resizeObserver = undefined;
+  }
+
+  private get effectiveVariant(): ResolvedWindVariant {
+    if (this.variant === WindVariant.auto) return this._autoVariant;
+    return this.variant;
```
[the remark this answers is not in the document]
CarelessCourage: Addressed. I added runtime validation for variant values and fall back to the computed auto variant if an invalid string is provided.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.stories.ts
```diff
@@ -93,6 +93,15 @@ const meta: Meta<typeof ObcWind> = {
...
       options: topVessels,
     },
+    variant: {
+      control: 'inline-radio',
+      options: [
+        WindVariant.auto,
+        WindVariant.small,
+        WindVariant.medium,
+        WindVariant.large,
+      ],
```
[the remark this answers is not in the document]
CarelessCourage: Addressed. I added a variant control description in Storybook documenting the auto thresholds and how the mode switches between variants.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
```diff
@@ -17,40 +36,126 @@ export class ObcWind extends LitElement {
...
+          tickmarksInside
+          .vessels=${vessels}
         ></obc-watch>
-        ${this.renderWindHistogram()}
+        ${this.renderWindArrow()}
+      </div>
+    `;
+  }
+
+  private renderWindArrow() {
```
talpitoo: Around this new `renderWindArrow()` and the dropped `windFromDirectionDeg`/`wind`/`windSymbolRadius`/`scaleWindIcon` props on `<obc-watch>`:

The repo's .github/instructions/watch-radial-instruments.instructions.md is explicit that **all circular rendering logic should live in `watch.ts`**, and that navigation instruments should be thin wrappers that *configure* `obc-watch`, not re-implement its rendering:

> 1. All rendering logic should live in `watch.ts` — it is the single source of truth for circular instrument rendering
> 3. If adding a new visual element, add it to `watch.ts` as a configurable option, not to individual instruments

This PR moves in the opposite direction:
- It stops using `obc-watch`'s built-in wind rendering (`wind`, `windFromDirectionDeg`, `windSymbolRadius`, `scaleWindIcon`).
- It re-implements an arrow overlay in `wind.ts` as an HTML/CSS-positioned `<div>` tree using `cqmin` units, pulling the SVG fragment directly out of `environmentSvgs`.

That has a few consequences:
1. Variant-aware sizing of the wind arrow now only works for `obc-wind` — `compass`, `heading`, etc. that use the `windFromDirectionDeg`/`scaleWindIcon` path on `obc-watch` won't benefit.
2. There are now two parallel code paths for "render a wind symbol at a radius", which will drift.
3. The CSS-driven layer doesn't share `obc-watch`'s `--scale` / `vector-effect="non-scaling-stroke"` strategy, so stroke widths and visual proportions will diverge from the rest of the instrument family.

Recommended approach (matches the existing pattern used for tickmarks, advices, vessels, etc.):

- Add a `windVariant` (or similar) configuration option to `obc-watch` that controls wind-symbol size, ring type, and inner placement.
- Keep `wind.ts` as a thin wrapper that:
  - owns `variant` + `auto` resolution via `ResizeObserver`,
  - passes the resolved variant down to `<obc-watch>` and to `renderWindHistogram`.
- Remove the HTML/CSS overlay arrow.

If you want to keep this PR scoped, an alternative is to pass the existing `windSymbolRadius`/`scaleWindIcon` per variant from `wind.ts` so the arrow keeps living inside `obc-watch`.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
talpitoo: @CarelessCourage just reiterating on this:

+1 to making `WIND_SMALL_MAX_PX` and `WIND_MEDIUM_MAX_PX` consumer-configurable — keep the current values as defaults. Suggested shape:

```ts
@property({type: Number}) smallVariantMaxPx = 96;
@property({type: Number}) mediumVariantMaxPx = 200;
```

Then `setAutoVariant` should call `resolveAutoWindVariant(min, this.smallVariantMaxPx, this.mediumVariantMaxPx)` (turn the helper into a pure function taking the thresholds).

Also: when these properties change in `updated()`, recompute `_autoVariant` so consumers don't have to resize the host to see the new thresholds take effect.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.css
```diff
@@ -15,3 +25,77 @@
...
+   * viewBox; allow it to paint past the SVG's edge instead of being
+   * clipped. */
+  overflow: visible;
+  /* Push the SVG out of the inner circle by its own height (100%) plus
+   * the inward margin, so it sits centred in the gap when --arrow-fill
+   * < 1, or flush against the outer circle when --arrow-fill = 1. */
+  transform: translateY(calc(100% + var(--arrow-margin)));
+}
+
+/* Per-variant tuning */
```
talpitoo: lines 87–101:

The three per-variant blocks set `--outer-pct: 80cqmin` identically and only differ by `--inner-frac`. Consider collapsing to:

```css
.wrapper .wind-arrow-outer {
  --outer-pct: 80cqmin;
  --inner-frac: 0.7; /* default = large */
}

.wrapper.variant-small .wind-arrow-outer  { --inner-frac: 0.1; }
.wrapper.variant-medium .wind-arrow-outer { --inner-frac: 0.4; }
```

(This is moot if comment `1` is taken and the arrow moves back into `obc-watch`.)

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
talpitoo: ...and another one for this method:

```ts
<svg class="wind-arrow" viewBox="0 0 24 24">${svg`${symbol}`}</svg>
```

If `environmentSvgs[…]` is already a Lit `TemplateResult` (which it is everywhere else it's consumed in `watch.ts`), wrapping it again with `` svg`${symbol}` `` is redundant — just interpolate `${symbol}` directly. If it's a raw string, this is unsafe and should go through `unsafeSVG` instead. Please pick one and drop the no-op tag.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
```diff
@@ -1,14 +1,33 @@
...
+}
+
+type ResolvedWindVariant =
+  | WindVariant.small
+  | WindVariant.medium
+  | WindVariant.large;
+
+const WIND_SMALL_MAX_PX = 96;
+const WIND_MEDIUM_MAX_PX = 200;
+const WIND_HISTOGRAM_MAX_RADIUS_SINGLE_RING = 159;
```
talpitoo: lines 27–28:

The new constants
```ts
const WIND_HISTOGRAM_MAX_RADIUS_SINGLE_RING = 159;
const WIND_HISTOGRAM_MAX_RADIUS_DOUBLE_RING = 112;
```
duplicate (and slightly drift from) the canonical ring radii exported from `watch.ts`:
```ts
export const OUTER_RING_RADIUS = 184;          // 368/2
const RING2_RADIUS              = 160;          // 320/2
const RING3_RADIUS              = 112;          // 224/2  ← same as DOUBLE_RING here
const RING3B_RADIUS             = 136;          // 272/2
```
112 lines up with `RING3_RADIUS`, but 159 is one pixel inside `RING2_RADIUS` (160) and isn't a documented constant. To avoid silent drift if the watch radii ever change, prefer importing the constants from `watch.ts` (or adding a named export there) rather than hard-coding new ones here.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
```diff
@@ -17,40 +36,126 @@ export class ObcWind extends LitElement {
...
+    if (!this._resizeObserver) {
+      this._resizeObserver = new ResizeObserver((entries) => {
+        const entry = entries[0];
+        if (!entry || this.variant !== WindVariant.auto) return;
+        const {width, height} = entry.contentRect;
+        this.setAutoVariant(width, height);
+      });
+    }
+
+    this._resizeObserver.disconnect();
```
talpitoo: Minor, but still:

```ts
this._resizeObserver.disconnect();
this._resizeObserver.observe(this);
this.setAutoVariant(this.clientWidth, this.clientHeight);
```

The `disconnect()` here is only needed because `syncResizeObserver` is called from both `connectedCallback` and `updated`. Since you already early-return when there's nothing to do (variant ≠ auto, no `ResizeObserver`), and the observer always observes the same element (`this`), it's safer to construct + observe exactly once and just no-op when already observing. The current `disconnect()`+`observe()` cycle on every `updated()` will trigger an extra `ResizeObserver` callback, which then calls `setAutoVariant` again. Not a bug, just churn.

talpitoo (changes requested):
Thanks for the contribution! The auto-variant logic, SSR guard, and value validation look good after the latest fixes. Two larger items remain before this can land:

1. **Architecture (must-fix per [`watch-radial-instruments.instructions.md`](.github/instructions/watch-radial-instruments.instructions.md))**: the wind arrow should be rendered by `obc-watch` (driven by a new variant option), not by an HTML/CSS overlay in `wind.ts`. Otherwise the wind icon rendering used by `compass`, `heading`, etc. drifts away from this one.
2. **Configurable breakpoints** (tibnor's comment): expose `smallVariantMaxPx` / `mediumVariantMaxPx` as properties.

Plus a few smaller items inline (CSS dedupe, double `svg` tag, snapshot baselines (i can generate them once all fixes are done and only Playwright snapshots are failing in the build), hard-coded histogram radii).

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
CarelessCourage: Good call. This is now done 🥳

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
CarelessCourage: Awesome 😊 Thats much better, good call. This change is now in place.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.css
CarelessCourage: I didn't even notice 😂 I was preparing to tweek them, but you're right. It's the same in all variants. Removed now. Good call.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
CarelessCourage: True. I removed the disconnect/re-observe cycle in `syncResizeObserver()` and now observe only once per active auto-mode lifecycle. So the extra ResizeObserver callback churn is gone while behavior stays the same.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
CarelessCourage: That's fair.

I originally had an issue where it seemed that the histogram was rendered on-top of the watch arrow which made things a bit complicated, but I think that might not be an issue after all. Additionally, since the arrow also needed some wind specific resizing behavior I figured it made more sense to extract it out to keep that behavior isolated to the wind element and not affect any other uses of the watch while still importing in the actual arrow SVG to keep at least that part unified. Especially considering I'm a bit unfamiliar with the codebase I didn't want to make changes that affect a lot of unrelated components if I can help it. 

But following your advice I'll have a deeper look at it and see what I can do to avoid rendering the arrow outside the watch. I'll let you know when that work is done. Thank you for your time. I appreciate all the great comments 😊

CarelessCourage: @talpitoo I've now moved over to using the arrow that's in the watch. This was totally the right move and it was actually much simpler than I thought. It's reduced the PR size as well and we no longer need any of that CSS logic I made. Sorry for overcomplicating it initially 😅

talpitoo: @CarelessCourage thank you, will take another look!

packages/openbridge-webcomponents/src/storybook-util.ts
talpitoo, line 1: @CarelessCourage the changes to `packages/openbridge-webcomponents/src/storybook-util.ts` (and the related `WindStoryArgs` alias added to `wind.stories.ts` in 27748e10 — "fix: fixes storybook type issues") aren't needed for this PR.

I tested locally by reverting both files to `develop` while keeping all of the wind responsiveness changes (including the new `variant: WindVariant` property and stories). Both `npm run typecheck` and `npm run lint:eslint` pass cleanly.

The reason it works: on `develop`, `wind.stories.ts` already used `Meta<typeof ObcWind>` together with `widthDecorator` and a `width` arg. TypeScript was happy because:
- `width: {control: ...}` in `argTypes` is fine (argTypes accepts arbitrary keys).
- `args: {width: 400}` is fine because `Meta`'s `args` is typed as `Partial<TArgs>` and Storybook tolerates extra keys.
- `widthDecorator`'s `context.args` was loose enough to accept any object.

Adding `variant: WindVariant` to `ObcWind` doesn't change that, because `variant` is a real component property — it doesn't widen the "extra args" surface.

Could you revert 27748e10 and push? Easiest:

```bash
git revert --no-edit 27748e10
npm run typecheck && npm run lint:eslint
git push
```

This keeps the PR scoped to the wind instrument and avoids touching a shared helper that every story file in the repo depends on. Happy to revisit the `widthDecorator` typing in a separate, dedicated PR if it's worth tightening.

talpitoo (changes requested):

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
```diff
@@ -17,40 +38,131 @@ export class ObcWind extends LitElement {
...
+        const {width, height} = entry.contentRect;
+        this.setAutoVariant(width, height);
+      });
+    }
+
+    if (!this._isObservingResize) {
+      this._resizeObserver.observe(this);
+      this._isObservingResize = true;
+    }
+    this.setAutoVariant(this.clientWidth, this.clientHeight);
```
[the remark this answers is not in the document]
CarelessCourage: Done 😊

packages/openbridge-webcomponents/src/storybook-util.ts
CarelessCourage: That makes sense 😊 Removed it now.

talpitoo (approved):
Re-reviewed after the latest commits — looks good. The architectural concern (duplicating wind rendering outside `obc-watch`) was elegantly resolved by reusing the existing `wind` / `windFromDirectionDeg` / `windSymbolRadius` / `scaleWindIcon` props on `<obc-watch>` and just forwarding per-variant values. That's why no `watch.ts` changes are needed — the watch already had the right knobs.

All previous review items are addressed:
- ✅ Variant-driven arrow rendering goes through `obc-watch` (no HTML/CSS overlay)
- ✅ `smallVariantMaxPx` / `mediumVariantMaxPx` exposed as properties (with min/max bounds for ordering robustness)
- ✅ Histogram radii use `innerRingRadiusFor(WatchCircleType.*)` from `watch.ts`
- ✅ ResizeObserver is gated on `variant === auto`, deferred to `firstUpdated`, and only observes once
- ✅ `storybook-util.ts` reverted
- ✅ Linux baselines updated
- ✅ `npm run typecheck` and `npm run lint:eslint` pass

Pre-existing oddities in `wind.ts` (hard-coded `viewBox="-200 -200 400 400"` for the histogram, no `getPadding()`) are out of scope here. Could be a follow-up to align with the `compass.ts` / `wind-propulsion.ts` viewBox calculation pattern.

LGTM from my side.

packages/openbridge-webcomponents/src/navigation-instruments/wind/wind.ts
```diff
@@ -17,40 +38,137 @@ export class ObcWind extends LitElement {
...
+      this._autoVariant = next;
+    }
+  }
+
+  private get effectiveVariant(): ResolvedWindVariant {
+    if (this.variant === WindVariant.auto) return this._autoVariant;
+    return isResolvedWindVariant(this.variant)
+      ? this.variant
+      : this._autoVariant;
+  }
```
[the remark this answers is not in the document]
talpitoo: Closing as minor — both edge cases are real but narrow:

- The stale-frame issue only manifests on dynamic `variant` switches; the common `auto`-from-start path is already handled by `firstUpdated`.
- The invalid-string path is only reachable via HTML typos and still renders a sensible fallback.

Happy to revisit as a follow-up if either becomes a real problem in practice.

ulrik-jo:
Looks very good. Thanks for your contribution, it is highly appreciated.

ulrik-jo (approved):

PR #881, opened 2026-05-15
fix: restrict published package to compiled output

AdamVig:
[//]: # (⇲)
`@oicl/openbridge-webcomponents@1.0.1` is 136 MB unpacked across
14,173 files. Two things are going wrong here.

First, test code is leaking into the library. Two build paths, each
picking up files it shouldn't.

The regular library build (`vite build`) feeds `src/**/*.ts` into
rollup, ignoring only `*.stories.ts`. That sweeps in `*.spec.ts`
files (which import `vitest` and `vitest-browser-lit`),
`src/storybook-util.ts` (which imports `@open-wc/lit-helpers`), and
`src/ar/_test-utils.ts` (a Storybook visual-test settle helper).
With `preserveModules: true` and an externals list that doesn't cover
test deps, rollup writes `vitest`, `@vitest`, `@open-wc`,
`magic-string`, `tinyrainbow`, etc. out under `dist/node_modules/`.
Extend the input glob's ignore list to also drop `*.spec.ts`,
`*.test.ts`, `src/storybook-util.ts`, and `src/ar/_test-utils.ts`,
and apply the same to `vite-plugin-dts` so it stops emitting `.d.ts`
for those files.

The bundle build (`vite build --mode bundle`) uses `bundle.ts`
generated by `script/generate-bundle-entry.ts`. #801 already excluded
`*.spec.ts` there, but `_test-utils.ts` still gets imported into
`bundle.ts` and inlined into the single-file bundle. Add
`**/_test-utils.ts` to the script's ignore list to close that gap.

Second, there's no `files` allowlist in `package.json`, so npm
publishes everything in the package directory:

- `__vis__/linux/` (40 MB), Playwright visual baselines
- `src/` (7 MB), already compiled into `dist/`
- `public/` (8 MB), demo assets
- `AR-test-image.png` shipped three times (24 MB) and `NotoSans.ttf`
  shipped three times (1.8 MB), once each under `bundle/`, `dist/`,
  and `public/`
- `companylogo-day.png`, used only by Storybook stories

Add an explicit `files` allowlist covering `dist/`, the standalone
bundle and its sourcemap, and the custom-elements manifest. The two
unused images get dropped via negation. Keep one copy of
`NotoSans.ttf` (in `dist/`) for consumers who want to self-host the
font with `@font-face`.

Also delete `public/vite.svg`. Leftover from `npm create vite`
scaffolding, nothing references it, no reason for it to keep
auto-copying into `dist/` on every build.

Result, measured locally: 7.2 MB tarball / 55.0 MB unpacked / 9,698
files, down from ~59 MB / 136 MB / 14,173.

[//]: # (⇱)

## Summary by CodeRabbit

* **Chores**
  * Tightened package publish settings to include only build outputs and necessary metadata, excluding two specific image assets for cleaner releases.
  * Refined build and declaration generation to consistently exclude story, spec/test, and test-utility source files from compiled bundles and type artifacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/881?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ulrik-jo:
Thanks for your contribution. It is highly appreciated, and we will have a look at this asap.

ulrik-jo:
Hi @AdamVig. We see that we also need a "full bundle"-version on npm. Hence, we will take over your PR and add the necessary actions to produce both your minimal bundle together with the full bundle. Hope this is fine by you.

PR #887, opened 2026-05-19
feat(number-input-field): format number

tibnor:
## Summary by CodeRabbit

* **New Features**
  * Number input now uses numeric values (NaN = empty), emits numeric input/change payloads, and supports configurable formatting (decimal/group separators, min/max fraction digits) plus a display override.

* **Bug Fixes**
  * Improved input validation, preservation/restoration of user edits (including trailing decimals), and focus/update behavior to avoid overwriting user-typed display.

* **Tests**
  * Added comprehensive tests for parsing, formatting, display behavior, and event semantics.

* **Documentation**
  * Storybook stories updated to treat value as numeric and reflect formatting controls.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/887?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #900, opened 2026-05-21
fix: text input align center

tibnor:


talpitoo:
Merged `develop` in. GitHub now shows **0 files changed** — every change on this branch has landed on develop independently, so this PR can be closed.

Where each piece went:

| This branch | On develop |
|---|---|
| `number-input-format.ts` + both spec files | arrived via #887 and #966 — those versions are supersets (locale-aware grouping separators, `isAllowedIntermediateInput`), and every test here already exists there |
| `updateCenterAlignedInputWidth` | byte-identical on develop, so the center-align fix is in |
| `showDivider` on `obc-icon-button` | on develop, header tag and `hide-divider` class included |
| `displayOverride`, `validationPattern` on `obc-number-input-field` | on develop |
| `handleInput` validation in `keyboard-numeric` | superseded — develop deleted `isValidValue` and moved the check into `obc-number-input-field`, which `effectiveValidationPattern` feeds |

That last one would no longer compile on its own: `isValidValue` does not exist on develop any more.

I pushed the merge rather than closing the PR, so the 0-diff is visible in the UI. Close whenever you are happy with that.

PR #903, opened 2026-05-22
feat(gauge-radial): add readout with sector-aware layout

### LINKED ISSUE #904: Gauge radial: add readout with sector-aware layout

anastasiyaBoyko:
issue #904 

## Summary by CodeRabbit

* **New Features**
  * Radial gauges: support 270°, 180°, 90° left/right sectors and optional readout with configurable label, unit, and decimal precision.
  * New layout controls: horizontal clipping and end-label placement (max/min) plus improved tick-label behavior.
  * Refined needle rendering and alignment for more consistent visuals.
  * Added Storybook scenarios demonstrating sectors, readouts and max/min label layouts.

* **Style**
  * CSS updates to adapt gauge and readout layouts per sector and center stacked readouts.

* **Documentation**
  * Added geometry inputs cheat‑sheet and guidance on clips vs. zoom.

- **`obc-watch` API extension:** adds `clipLeft` / `clipRight` to `obc-watch` (horizontal counterparts of the existing `clipTop` / `clipBottom`, for quadrant / 90° sectors). Despite the "gauge-radial" title, this is a real addition to the shared renderer's public API.
- **Needle rendering:** the needle is drawn at a single, non-scaled size across all sectors (270° / 180° / 90°); the earlier proportional needle-scaling path was removed. Confirmed with design that this simpler needle is intended for every sector.
- **`instrument-radial` viewBox:** the un-zoomed `448` is now the named constant `WATCH_DEFAULT_VIEWBOX`, coupled to `obc-watch`'s default padding (`(176 + 48) * 2`); update it if that default padding changes.


talpitoo (changes requested):
@anastasiyaBoyko i smoketested it and found a few issues:
- https://openbridge-next-storybook--903-feat-gauge-radial-reado-c7yb57qf.web.app/?path=/story/instruments-gauge-radial--sector-90-left and https://openbridge-next-storybook--903-feat-gauge-radial-reado-c7yb57qf.web.app/?path=/story/instruments-gauge-radial--sector-90-right and all other sector 90 gauges are 'thinner', their watch face is thinner. I think that should not be the case, that regardless of the angle, the watch face thickness should remain the same.

I haven't looked at Figma, neither do i know the specs, i just comment it based on the work i did on similar 'radial' instruments.

talpitoo (changes requested):
@anastasiyaBoyko i noticed the combination of these properties with `tickmarksInside` lead to a broken component: https://openbridge-next-storybook--903-feat-gauge-radial-add-r-00pxam30.web.app/?path=/story/instruments-gauge-radial--sector-90-left-with-readout&args=state:loading;tickmarksInside:!true

Image

talpitoo:
The snapshots show an updated needle:

Screenshot from 2026-06-01 17-17-52
Screenshot from 2026-06-01 17-17-57

Is that a new design in Figma (haven't checked yet) or accident?

talpitoo:
**File:** `watch.ts` (around `_bandRadius`) + `building-blocks/instrument-radial/instrument-radial.ts`

`tickmarksInside` still produces a broken component on 90° sectors (see screenshot in a previous comment).

The band shifts inward but the tickmarks/labels stay at their additive radii, so they end up floating outside the ring.

The JSDoc on `_bandRadius` already acknowledges this:

> Tickmark and mask radii stay additive on purpose, so proportional scaling assumes outside tickmarks — `tickmarksInside` would misalign with the band.

...but the preferred way would be to make tickmark/mask radii respect `_bandRadius` too, so the two layers stay coherent (`tickmark.ts` and `advice.ts` already accept a radius offset: a multiplicative variant would fit naturally).

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.css
```diff
@@ -0,0 +1,81 @@
+:host {
+  display: block;
+  width: 100%;
+  height: 100%;
+}
+
+:host([sector="90-left"]),
```
talpitoo, line 10: Hard-coded `200×200` host size breaks responsive sizing convention

**File:** gauge-radial.css

```css
:host([sector="90-left"]),
:host([sector="90-right"]) {
  width: 200px;
  height: 200px;
}
```

This contradicts what every other watch-based instrument does (see compass.ts → `:host { display: block; width: 100%; height: 100%; }` and the `widthDecorator` story pattern) — the host fills its parent and the parent decides the size. Hard-coding `200px` here means:

- The Storybook `width` arg no longer affects the 90° stories (try the range slider in [Sector 90 Right](https://openbridge-next-storybook--903-feat-gauge-radial-add-r-00pxam30.web.app/?path=/story/instruments-gauge-radial--sector-90-right) — nothing happens).
- Consumers who want a larger 90° gauge in a real layout silently get clipped at 200px.
- It breaks the `obc-component-size-*` size-class story convention (sizing should flow from the parent, only text scales).

Drop the override; if the corner-readout layout assumes a minimum size, use `min-width`/`min-height` instead, or solve it with `aspect-ratio` so the gauge stays square at whatever width it's given.

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.css
```diff
@@ -663,6 +667,7 @@
   @mixin font-instrument-label-box;
   color: var(--element-neutral-color);
   @mixin font-instrument-label;
+  font-weight: var(--global-typography-instrument-label-font-weight);
```
talpitoo:
```css
.value-unit {
  ...
  @mixin font-instrument-label;
+ font-weight: var(--global-typography-instrument-label-font-weight);
  font-size: var(--global-typography-instrument-unit-font-size);
  ...
}
```

This affects **every** `obc-readout` in the library, not just gauge-radial.. Two possibilities:

- It's an intentional design fix: please split it into its own commit + PR (or at least a separate commit here) with a one-line rationale, and ping the readout owners.
- It's a workaround to make gauge-radial look right: please solve it locally (scoped class or `::part`) instead of changing every readout in the library.

Either way, this needs to be called out explicitly in the PR description. Right now it's hidden inside "sector-aware layout".

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.css
```diff
@@ -0,0 +1,81 @@
...
+  &.sector-90-left {
+    .gauge-readout-meta {
+      left: auto;
+      right: 0;
+      bottom: 0;
+      top: auto;
+      transform: none;
+    }
+
+    obc-readout.gauge-readout-meta::part(value-icon-container) {
```
talpitoo: Dead CSS selectors and an unused exposed `part`, also in `readout-setpoint.ts`

The 90° corner-readout block targets parts that **do not exist** on `obc-readout`:

```css
obc-readout.gauge-readout-meta::part(value-icon-container) { display: none; }   /* ❌ no such part */
obc-readout.gauge-readout-meta::part(meta-wrapper),                              /* ❌ no such part */
obc-readout.gauge-readout-meta::part(instrument-label-unit-container) { ... }    /* ❌ no such part */
```

These rules silently no-op. Either expose the matching parts on `obc-readout` (and update the readout JSDoc), or remove the dead selectors - please don't leave them.

Related: the change to readout-setpoint.ts adds `part="icon-container"` to two `<div>`s, but **gauge-radial doesn't use `obc-readout-setpoint` anywhere** (only `obc-readout`). The `part` addition is dead code for this PR. Please drop it, or move it to a separate PR that documents *why* the `obc-readout-setpoint` part API needs that hook.

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
```diff
@@ -99,46 +120,185 @@ export class ObcGaugeRadial extends SetpointMixin(LitElement) {
...
-      return -135;
+      return start;
     }
 
-    return ((v - this.minValue) / span) * 270 - 135;
+    return ((v - this.minValue) / span) * sweep + start;
+  };
+
+  private renderCornerReadout(): TemplateResult {
+    return html`
```
talpitoo, line 207: DRY (minor): collapse the four `render…Readout` methods

`renderCornerReadout`, `renderBottomStackReadout`, `renderCenterValueReadout`, `renderLabelOnlyMetaReadout` all repeat the same ~10 property bindings. A single helper that takes a small `overrides` object (variant, alignment, class, optional label-only flag) would halve the surface area and make the sector/type matrix in `renderReadouts()` easier to follow.

Not a blocker, just for readability.

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.css
```diff
@@ -0,0 +1,81 @@
...
+
+  .gauge-readout-value {
+    position: absolute;
+    left: 50%;
+    top: 50%;
+    transform: translate(-50%, -50%);
+    z-index: 1;
+  }
+
+  .gauge-readout-meta {
```
talpitoo, line 43: Magic-number positioning of the meta readout

```css
.gauge-readout-meta            { top: 72%; }
&.type-needle .gauge-readout-meta { top: 63%; }
&.sector-180  .gauge-readout-meta { top: 60%; }
```

These percentages will quietly drift if anyone touches the tickmark padding, viewBox, or scale ring sizes in `watch.ts`. Since the inner radii are all known SVG-space constants (`OUTER_RING_RADIUS`, `RING2_RADIUS`, …), it'd be more robust to either (a) anchor the readout to the SVG via `anchor-name`/`position-anchor` the way instrument-radial.ts already does for `--watch`, or (b) drive these offsets from a single calculated CSS variable rather than four hand-tuned percentages.

talpitoo (changes requested):
**Summary of required changes for merge:**
1. Fix or hard-guard `tickmarksInside` on 90° sectors.
2. Remove the hard-coded `200×200` host size.
3. Split / justify the global `readout.css` `font-weight` change.
4. Remove dead `::part(...)` selectors and the unused `part="icon-container"` addition in readout-setpoint.ts.

The rest are nice-to-haves but should still be discussed.

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
```diff
@@ -99,46 +121,200 @@ export class ObcGaugeRadial extends SetpointMixin(LitElement) {
...
+          .getAngle=${this.getAngle}
+          .showLabels=${this.showLabels}
+          .primaryTickmarkInterval=${this.primaryTickmarkInterval}
+          .secondaryTickmarkInterval=${this.secondaryTickmarkInterval}
+          .tertiaryTickmarkInterval=${this.tertiaryTickmarkInterval}
+          .type=${this.type}
+          .needleType=${this.type}
+          .tickmarksInside=${this.tickmarksInside}
+          .tickmarkStyle=${this.tickmarkStyle}
+          .advices=${this.isSector90 ? [] : this.advices}
```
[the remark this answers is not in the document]
anastasiyaBoyko: The band/tickmark mismatch what describing came from the proportional band scaling (preserveBandProportion), where the band used a multiplicative radius while tickmarks/labels stayed additive. That mode was removed in this PR — _bandRadius is now plain additive (base + _rOff), identical to the tickmark and mask radii, and the 90° sectors render at natural scale via a crop (radiusOffset = 0), so the band isn't shifted inward at all. As a result the inside tickmarks/labels sit on the same radii as the band and stay aligned (verified on the 90° sectors with tickmarksInside enabled). The note that flagged this was removed along with the proportional path, so no coercion or separate radius fix is needed.

packages/openbridge-webcomponents/src/navigation-instruments/watch/tickmark.ts
```diff
@@ -135,8 +142,21 @@ function textSvg(
...
+  // 180° gauge: place the horizontal end labels (0/100) off the dead-center
+  // tick — below for outside ticks, lifted for inside — and inward (the wider
+  // "100" a touch more), so they tuck under the band ends.
+  if (endLabelsBelow && Math.abs(Math.cos(radHoriz)) < 1e-6) {
+    const sin = Math.sin(radHoriz);
+    const inward = inside ? (sin > 0 ? 11 : 6) / scale : 14 / scale;
+    const x = sin * (textRadius - inward);
+    const y = inside ? -(6 / scale) : 12 / scale;
+    return svg`<text x=${x} y=${y} class="label bottom ${inside ? 'inside' : ''}">${text}</text>`;
+  }
```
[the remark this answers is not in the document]
anastasiyaBoyko: .label.bottom (and .label.top) intentionally inherit the base .label rule, which is text-anchor: middle; dominant-baseline: central — i.e. centered — and that's exactly the alignment these end labels need. Only .left/.right override the anchor (end/start); top/bottom are position markers that deliberately fall back to the centered base, so no dedicated rule is required. This isn't new to this change either: textSvg already emits class="label bottom" for angle === 180 and class="label top" for angle === 0, both relying on the same base styling. So the class is styled (via .label) and not dead — adding a .label.bottom rule would just re-declare the inherited defaults.

anastasiyaBoyko:
> **File:** `watch.ts` (around `_bandRadius`) + `building-blocks/instrument-radial/instrument-radial.ts`
> 
> `tickmarksInside` still produces a broken component on 90° sectors (see screenshot in a previous comment).
> 
> The band shifts inward but the tickmarks/labels stay at their additive radii, so they end up floating outside the ring.
> 
> The JSDoc on `_bandRadius` already acknowledges this:
> 
> > Tickmark and mask radii stay additive on purpose, so proportional scaling assumes outside tickmarks — `tickmarksInside` would misalign with the band.
> 
> ...but the preferred way would be to make tickmark/mask radii respect `_bandRadius` too, so the two layers stay coherent (`tickmark.ts` and `advice.ts` already accept a radius offset: a multiplicative variant would fit naturally).

@talpitoo 
Resolved by removing the proportional band scaling entirely rather than making the tickmark/mask radii multiplicative.
Why this path instead of the suggested one: the proportional scaling (preserveBandProportion) only existed to keep the band thickness right while the 90° sectors were zoom-enlarged to fill their box. In PR the gauge no longer zoom-enlarges — to make every sector share the same dial scale (per the Figma spec), each sector now renders at natural scale and just crops a different window of the same 448 box (radiusOffset = 0). Once there's no enlargement, there's nothing to scale proportionally, so preserveBandProportion became dead code and the multiplicative tickmark/mask variant would have been solving a problem that no longer exists. Keeping the multiplicative path would also mean permanently keeping band + tickmark + mask + needle radii in sync by a factor. dropping it lets all of them stay on one plain additive formula.
Concretely: _bandRadius is now base + _rOff, identical to the tickmark and mask radii, so the band and inside tickmarks/labels sit on the same radii and stay aligned (verified on the 90° sectors with tickmarksInside).The _bandRadius JSDoc note about the mismatch was removed along with the proportional path.

anastasiyaBoyko:
> The snapshots show an updated needle:
> 
> Screenshot from 2026-06-01 17-17-52 Screenshot from 2026-06-01 17-17-57
> Is that a new design in Figma (haven't checked yet) or accident?

@talpitoo 
its related to figma
Знімок екрана 2026-06-03 о 12 42 59

packages/openbridge-webcomponents/src/navigation-instruments/readout/readout.css
anastasiyaBoyko: removed , as it solved in separate PR

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.css
anastasiyaBoyko: Done — dropped the fixed 200×200 and solved it with aspect-ratio as suggested. The base host keeps the display: block; width: 100%; height: 100% convention, and each sector just pins the aspect of its crop so it stays the right shape at whatever width the parent gives

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.css
anastasiyaBoyko: Done on both:
Dead ::part selectors — removed all three (value-icon-container, meta-wrapper, instrument-label-unit-container),none of those parts exist on obc-readout, so they were silent no-ops
part="icon-container" in readout-setpoint.ts — removed from both div. The gauge only uses obc-readout, so that hook was dead for this PR.
One thing to flag for transparency: the 90° block now has a single ::part(value-wrapper) rule (left-aligns the value under the label for the bottom-left corner readout). Unlike the ones above, value-wrapper is a real exposed part on obc-readout (part="value-wrapper" in readout.ts), so that one targets an existing part and isn't dead — it was needed because the enhanced variant right-aligns the value by default.

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
anastasiyaBoyko: Done — collapsed methods into a single renderReadout(opts) helper that takes the overrides you described: { className, variant, alignment, withValue, withMeta, labelOnly }. The shared <obc-readout> bindings now live in one place, and renderReadouts() reads as the sector/type matrix calling the helper

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.css
anastasiyaBoyko: Im looked at both a and b and neither maps cleanly onto the shadow-DOM structure, so I kept the percentages but made the coupling explicit instead of silent. Reasoning:
(a) anchor-name/position-anchor — --watch is declared inside instrument-radial's shadow root, while the readout lives in gauge-radial's shadow root. Anchor names are tree-scoped, so the readout can't reference --watch across that boundary. The only anchor reachable from the readout's scope is the <obc-instrument-radial> host, but it's stretched to 100% of the root, so anchoring to it just reproduces the same percentage — it carries no information about where the arc actually sits inside the viewBox.
(b) single calculated variable — the percentages aren't equal to any single ring constant. e.g. the 270 value (72%) maps to an SVG radius of ~99, i.e. roughly centered in the arc's lower opening — an aesthetic placement, not a geometric landmark. A calc() from the ring radii would just re-encode the same tuned number as a tuned coefficient. And the viewBox that actually drives the mapping is computed in JS inside instrument-radial (and now differs per sector), so it isn't available to the sibling readout via CSS inheritance.
What was done instead: pulled the offsets into named CSS variables with a comment documenting that they track the watch geometry and must be re-checked if the ring radii or the 448 box change. That addresses the real concern — "quietly drift" — by making the dependency explicit rather than silent. 
A fully geometry-driven version would mean exposing the live arc frame out of the shared instrument-radial building block (+ runtime sizing to handle viewBox letterboxing) — disproportionate to the risk here, but happy to land it as a focused follow-up if you'd prefer

anastasiyaBoyko:
> **Summary of required changes for merge:**
> 
> 1. Fix or hard-guard `tickmarksInside` on 90° sectors.
> 2. Remove the hard-coded `200×200` host size.
> 3. Split / justify the global `readout.css` `font-weight` change.
> 4. Remove dead `::part(...)` selectors and the unused `part="icon-container"` addition in readout-setpoint.ts.
> 
> The rest are nice-to-haves but should still be discussed.

@talpitoo responded separate in every note comment

anastasiyaBoyko:
@talpitoo 
Small scope note
The original task was to integrate the readout into obc-gauge-radial, which at that point only implemented the standard 270° sector. But figma spec was updated to add the 180° and 90-left / 90-right sectors, so the readout had to work across all of them — and getting those new sectors to match the spec (same dial scale across sectors, per-sector readout/label placement) required a set of changes beyond the original scope

talpitoo:
> @talpitoo Small scope note The original task was to integrate the readout into obc-gauge-radial, which at that point only implemented the standard 270° sector. But figma spec was updated to add the 180° and 90-left / 90-right sectors, so the readout had to work across all of them — and getting those new sectors to match the spec (same dial scale across sectors, per-sector readout/label placement) required a set of changes beyond the original scope

@anastasiyaBoyko thank you, i'll double-check your individual replies, the visual testing already looks good!
My only remaining (visual) question is that the labels (angles) are sometimes 'outside' of the gauge, sometime 'below/inline', see the attached screenshots. I feel that this should be triple-checked :-) with the designers. I see this in Figma too, just haven't checked whether it is a property/setting or hardcoded.

Screenshot from 2026-06-03 13-07-30
Screenshot from 2026-06-03 13-07-41
Screenshot from 2026-06-03 13-07-52
Screenshot from 2026-06-03 13-08-02

anastasiyaBoyko:
> > @talpitoo Small scope note The original task was to integrate the readout into obc-gauge-radial, which at that point only implemented the standard 270° sector. But figma spec was updated to add the 180° and 90-left / 90-right sectors, so the readout had to work across all of them — and getting those new sectors to match the spec (same dial scale across sectors, per-sector readout/label placement) required a set of changes beyond the original scope
> 
> @anastasiyaBoyko thank you, i'll double-check your individual replies, the visual testing already looks good! My only remaining (visual) question is that the labels (angles) are sometimes 'outside' of the gauge, sometime 'below/inline', see the attached screenshots. I feel that this should be triple-checked :-) with the designers. I see this in Figma too, just haven't checked whether it is a property/setting or hardcoded.
> 
> Screenshot from 2026-06-03 13-07-30 Screenshot from 2026-06-03 13-07-41 Screenshot from 2026-06-03 13-07-52 Screenshot from 2026-06-03 13-08-02

@talpitoo 
Yes, this is intentional, not a one-off hardcode: it's driven by an endLabelsBelow flag the gauge sets only for the 180° sector, where the horizontal end labels (0/100) tuck under the band ends while the rest stay radial/outside — matching the updated Figma for that sector

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.css
talpitoo: @anastasiyaBoyko good, i am going to resolve this comment, however please consider adding this into a TODO comment or similar for an eventual future refactoring: 

>  One thing to flag for transparency: the 90° block now has a single ::part(value-wrapper) rule (left-aligns the value under the label for the bottom-left corner readout). Unlike the ones above, value-wrapper is a real exposed part on obc-readout (part="value-wrapper" in readout.ts), so that one targets an existing part and isn't dead — it was needed because the enhanced variant right-aligns the value by default.

packages/openbridge-webcomponents/__vis__/linux/__baselines__/building-blocks/instrument-radial/instrument-radial.stories.ts/enhanced-needle-auto.png
talpitoo, line 1: @anastasiyaBoyko you've said this was confirmed with design that the simpler needle is intended across all sectors. I think it should me mentioned in the PR description (in the final squash commit message body).

packages/openbridge-webcomponents/src/navigation-instruments/watch/tickmark.ts
```diff
@@ -49,6 +49,7 @@ export function tickmark(
     maxDigits,
     color,
     radiusOffset = 0,
+    endLabelsBelow = false,
```
talpitoo: ### `endLabelsBelow` — name and per-side label-width literals
**Severity:** Flag-only. Works for the shipped sectors; just needs a paper trail.

---

Two things still worth doing as soft follow-ups, not blockers:

- **Rename for intent.** `endLabelsBelow` describes mechanism (place the label below the tick) but not the concept (the new design-language **Max-min** mode). Consider `endLabelsMaxMin` / `maxMinEndLabels` / `endLabelMode: 'beside' | 'maxMin'`. If you'd rather not churn the API, please at least drop the design term **"Max-min"** explicitly into the JSDoc:
  ```ts
  /**
   * Place the horizontal end labels (±90°, e.g. min/max) below the tick instead
   * of beside it. This is the "Max-min" label placement from the radial label
   * model (External / Internal / Max-min) — see PR #903 / design discussion.
   * Currently only used by the 180° sector of `obc-gauge-radial`.
   */
  ```
  That way future contributors can find it by name.

- **Asymmetric inward inset (`(sin > 0 ? 11 : 6) / scale`)** still encodes "the right label happens to be wider than the left" — true for `0..100` but not for, say, `-100..1000`. With the design intent now clear, this is the place to either:
  - drive the inset from the actual rendered text length (you have `maxDigits` upstream in `watch.ts:textRadius`), or
  - just split into `leftInset` / `rightInset` derived from the label `text.length` instead of hard-coded `11` / `6`.

  Leaving a `// TODO: derive from label width` comment + a story with `minValue=-50, maxValue=999` (which would visibly fail today) would be enough to make the future fix concrete.

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
```diff
@@ -50,6 +67,11 @@ export interface GaugeRadialAdvice {
  * - Provide `primaryTickmarkInterval` and `secondaryTickmarkInterval` to
  *   control tickmark density.
  * - Enable `showLabels` to show numeric labels at primary tickmarks.
+ * - Enable `showReadout` with optional `label` and `unit`. Layout depends on `sector`
+ *   and `type`: **270** / **180** filled/bar — centered value plus bottom
```
talpitoo: **Severity:** minor

---

JSDoc currently says:
> **270** / **180** filled/bar — centered value plus bottom label/unit row

But gauge-radial.ts uses **two** readouts for 270 (value at center + label-only meta lower) and **one** centered stack for 180. The visual result is similar but the DOM/parts differ — anyone extending or theming via `::part` will be surprised.

Fix is one bullet split:
- **270** filled/bar — value at center + label/unit row in the bottom gap.
- **180** filled/bar — single centered stack (value + label/unit) at the bottom.

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
talpitoo, line 307: **Severity:** minor

---

`.advices=${this.isSector90 ? [] : this.advices}`

Already documented in the class-level JSDoc — good. The remaining nit is **discoverability**: consumers reading the property declaration won't see it. Add a one-liner on the property itself:

```ts
/** Caution/alert arcs. Ignored on `sector: 90-left` / `90-right`. */
@property({type: Array, attribute: false}) advices: GaugeRadialAdvice[] = [];
```

Optionally a `console.warn` once in dev when `sector` is 90 and `advices.length > 0`. Up to you — I'd take just the JSDoc.

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
talpitoo, line 1: **Severity:** minor

---

gauge-radial.ts

`labelOnly: true` already implies "no value". Either:
- infer `withValue = !labelOnly` inside the helper and drop `withValue` from the call site, or
- drop the explicit `withValue: false` at the call site since `labelOnly` covers it.

One-line cleanup. Not a blocker.

packages/openbridge-webcomponents/src/navigation-instruments/watch/tickmark.ts
```diff
@@ -87,7 +89,9 @@ export function tickmark(
     innerRadius = 328 / 2 + rOff;
     outerRadius = 336 / 2 + rOff;
   } else {
-    return [textSvg(text ?? '', angle, inside, scale, textRadius)];
+    return [
+      textSvg(text ?? '', angle, inside, scale, textRadius, endLabelsBelow),
```
talpitoo: **Severity:** minor
### `textSvg` positional-param creep

---

`textSvg(text, angle, inside, scale, textRadius, endLabelsBelow = false)` is now 6 positional params, two of them booleans. Same pattern as the outer `tickmark()` itself — please convert to an options object the next time you touch this file. Will compose better when more end-label modes appear.

packages/openbridge-webcomponents/__vis__/linux/__baselines__/building-blocks/instrument-radial/instrument-radial.stories.ts/enhanced-needle-auto.png
anastasiyaBoyko: @talpitoo added note to PR description

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
```diff
@@ -99,46 +121,200 @@ export class ObcGaugeRadial extends SetpointMixin(LitElement) {
...
+  }
+
+  private get isSector90(): boolean {
+    return (
+      this.sector === GaugeRadialSector.deg90Left ||
+      this.sector === GaugeRadialSector.deg90Right
+    );
+  }
+
+  getAngle = (v: number): number => {
```
talpitoo, line 181: **Severity:** minor
`getAngle` arrow-property

---

Still an arrow without a comment. One-liner is enough:
```ts
// Arrow form so `this` binds when passed as `.getAngle=${this.getAngle}`
// to <obc-instrument-radial>.
```
Otherwise someone will "fix it" to a method and quietly break the bound `this`.

packages/openbridge-webcomponents/src/navigation-instruments/watch/watch.ts
talpitoo, line 1: @anastasiyaBoyko the following are all minor issues, falling into the 'future proofing' category becase the changes in `watch.ts`, `instrument-radial.ts` and `arc-frame.ts` would affect other radial instruments as well, hence sensitive. Still, i'll leave them here. If you find, you could address them, if not, please ping me to create a follow-up/minor/maintenance issue for myself 🙏 .

---

## 🆕 Future-proofing flags on `watch.ts` / instrument-radial.ts / `arc-frame.ts`

These are the parts you specifically called out as load-bearing. None are blockers for this PR; they're notes to either embed as comments now or capture as follow-up issues so the next person touching this surface doesn't have to re-derive it.

### F1. `clipLeft` / `clipRight` × `zoomToFitArc` are mutually exclusive — silently

instrument-radial.ts and watch.ts

Both `instrument-radial` and `watch` now zero out **all four clips** when `zoomToFitArc` is true:

```ts
.clipTop=${this.zoomToFitArc ? 0 : this.clipTop}
.clipBottom=${this.zoomToFitArc ? 0 : this.clipBottom}
.clipLeft=${this.zoomToFitArc ? 0 : this.clipLeft}
.clipRight=${this.zoomToFitArc ? 0 : this.clipRight}
```

That's the right call (the zoom path computes its own viewBox), but it's invisible to consumers. Two existing instruments that mix clip + zoom — `rudder` (`clipTop=40`) and `compass-sector` — already rely on this behavior. Please add a one-liner on each of the four `clip*` properties in `watch.ts`:

```ts
/** Top clip, % of height. Ignored when `zoomToFitArc` is true. */
@property({type: Number}) clipTop: number = 0;
```

And consider folding this into watch-radial-instruments.instructions.md § "Clipping for Partial Gauges":

> **Note:** `clip{Top,Bottom,Left,Right}` are mutually exclusive with `zoomToFitArc`. When zoom is on, the viewBox is derived from `computeZoomToFitArcFrame()` and clips are ignored.

### F2. `clipLeft` / `clipRight` are new public API on `obc-watch`

Adding `clipLeft` and `clipRight` to `obc-watch` is a generally useful change, but it's now part of the public renderer API and every other consumer can set them. Two soft asks:

- Mention this in the PR description / changelog — the PR title is "gauge-radial" but `obc-watch` got a real API extension.
- Add a brief usage hint to the JSDoc (e.g. "horizontal counterpart of `clipTop`/`clipBottom`; useful for quadrant sectors") so future consumers don't re-invent it.

### F3. The shared 448 box is now load-bearing in two places

instrument-radial.ts hard-codes `full = 448`; watch.ts computes `(176 + padding) * 2` (which for `padding=48` is also 448). The `gauge-radial.css` aspect ratio `448 / 251` also bakes the same constant.

These three coincidentally line up at `padding=48` but they are derived from different code paths. Future risk: if anyone changes the `padding` of `instrument-radial` from 48, the `448` in the `else` branch of `instrument-radial.ts:render` becomes wrong (watch and overlay viewBox will silently disagree by twice the padding delta).

Two ways to harden, pick one:

- Extract a `INSTRUMENT_RADIAL_PADDING = 48` (or `INSTRUMENT_RADIAL_VIEWBOX = 448`) constant and reuse it in both files. Best.
- Or at least a comment in instrument-radial.ts next to `const full = 448`: `// must match obc-watch's (176 + padding) * 2 with padding=48 below`.

### F4. `computeZoomToFitArcFrame` is now invoked from two places with subtly different lifecycles

After this PR the same function runs both inside `obc-watch` (when `arcFrame` is undefined) **and** inside `obc-instrument-radial` (which then passes the precomputed frame down via `.arcFrame`). `instrument-radial` always passes `arcFrame`, so `watch` skips its own computation — good.

The flag here is the ObcWatch.arcFrame JSDoc: it doesn't currently spell out **"if you pass `arcFrame`, you own keeping it in sync with `areas` / `watchCircleType` — `obc-watch` will not recompute"**. Worth one extra sentence so a future consumer doesn't pass a stale frame.

### F5. `radiusOffset` is "additive everywhere now" — please document the invariant

watch.ts — `_bandRadius(base) => base + _rOff`

The whole reason the proportional path was removed is that "additive everywhere" is the invariant that keeps band thickness, tickmark anchoring, label `textRadius`, and `roundedArch` masks coherent under zoom. Right now that's encoded as "every site uses `_bandRadius`". A future contributor adding a new ring or decoration might add `+ _rOff` in one place and forget another.

Cheap fix: in `watch.ts`, above `_bandRadius`, add a short comment:

```ts
/**
 * INVARIANT: every band-edge radius under zoom must go through _bandRadius()
 * (additive offset). Mixing additive and multiplicative offsets in the same
 * watch face misaligns tickmarks, labels, advice masks and roundedArch arcs.
 * Earlier proportional-scaling experiments were removed for this reason — see
 * PR #903.
 */
private _bandRadius(base: number): number { return base + this._rOff; }
```

### F6. `tickFadeAngle`, `arcFrame`, `clip*`, `zoomToFitArc`, `padding` matrix is getting hard to reason about

This isn't a code change — it's a flag. After this PR, `obc-watch` exposes the following partially-overlapping geometry inputs:

| Property | Affects |
|---|---|
| `padding` | base viewBox size in the un-zoomed path |
| `clipTop/Bottom/Left/Right` | viewBox window in the un-zoomed path; **ignored** when `zoomToFitArc` |
| `zoomToFitArc` | swaps to `computeZoomToFitArcFrame()` path; band radii get `_rOff` |
| `arcFrame` | externally-precomputed zoom frame; only used if `zoomToFitArc` is on |
| `endLabelsBelow` | Max-min label placement at horizontal ends |
| `tickmarksInside` | label radius (now goes through `_bandRadius`) |
| `tickFadeAngle` | (pre-existing) tickmark fade-out near arc edges |

Pitch-roll already taps into `computeZoomToFitArcFrame` + `shiftArcFrameToOuterEdge` and assumes specific combinations. Worth adding (in a follow-up PR, not this one) a short "Geometry inputs cheat-sheet" section to watch-radial-instruments.instructions.md describing which combinations are validated and which are undefined. The Slack design model (External / Internal / Max-min) belongs there too.

### F7. Inside-label `textRadius` now uses `_bandRadius` — quick sanity check needed

watch.ts `textRadius` was changed from `inner/outer + rOff` to `_bandRadius(inner/outer)`. Functionally identical today (because `_bandRadius` is currently additive), but it now silently couples to whatever `_bandRadius` becomes in the future. If you ever reintroduce a non-additive mode, inside labels will follow it for free — which may or may not be what you want. Worth a one-line comment that this coupling is intentional.

talpitoo (changes requested):
The architectural shape of this PR is good — the additive-only `_bandRadius`, the viewBox-crop sector model, the design-aligned label modes, and the precomputed `arcFrame` handoff are all moves in the right direction. The remaining items are mostly about leaving enough breadcrumbs so the next person editing `watch.ts` doesn't accidentally break one of the invariants that this PR established.

For context: my previous review pass was primarily visual — confirming that sector layouts, needle, readout positioning, and tickmark behavior all looked correct across the variants. This round is mostly AI-assisted code-cleanliness review (JSDoc clarity, naming, API surface, future-proofing comments around the shared `obc-watch` / `obc-instrument-radial` machinery). Treat them as suggestions, not blockers — happy to land most of them as a follow-up if that keeps this PR focused.

packages/openbridge-webcomponents/__vis__/linux/__baselines__/building-blocks/instrument-radial/instrument-radial.stories.ts/enhanced-needle-auto.png
talpitoo: @anastasiyaBoyko please add it to the final squash message as well so it won't be forgotten. Resolving this now.

packages/openbridge-webcomponents/src/navigation-instruments/watch/tickmark.ts
anastasiyaBoyko: @talpitoo 
Renamed for intent. Took the full rename rather than the JSDoc-only minimum: endLabelsBelow → endLabelsMaxMin across watch / instrument-radial / gauge-radial / tickmark. Added the JSDoc you suggested, documenting the Max-min placement and the External / Internal / Max-min label model, noting it's currently only used by the 180° sector of obc-gauge-radial.
Width-derived inset: replaced the hard-coded (sin > 0 ? 11 : 6) with an inset driven by the actual label width: inside ? (6 + (text.length - 1) * 2.5) / scale : 14 / scale. It no longer assumes the right label is the wider one — it reproduces the old 6/11 values for "0"/"100" and generalizes to any range. Added a Sector180WideEndLabels regression story to lock this in. (I used -250 / 1000 with a sparse primaryTickmarkInterval rather than the exact -50..999 — same intent of wide labels on both ends, but with grid-aligned min/max so the end labels don't collide with an adjacent 0 tick.)

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
anastasiyaBoyko: @talpitoo 
Split the bullet. JSDoc now documents the two layouts distinctly:
270 filled/bar — value at the center + a label-only meta row in the bottom gap (two separate readouts).
180 filled/bar — a single centered stack (value + label/unit) at the bottom.
So the differing DOM/parts is explicit for anyone theming via ::part.

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
anastasiyaBoyko: @talpitoo 
Added the one-liner on the property declaration,
skipped the console.warn per your preference.

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
anastasiyaBoyko: @talpitoo 
withValue is now derived inside renderReadout (const withValue = !labelOnly) and the withValue param + the explicit withValue: false at the call site — are gone. Behavior is unchanged

packages/openbridge-webcomponents/src/navigation-instruments/watch/tickmark.ts
anastasiyaBoyko: @talpitoo 
Done — converted to an options object (matching tickmark()):
textSvg now takes (text, { angle, inside, scale, textRadius, endLabelsMaxMin }) instead of 6 positional params, mirroring the outer tickmark() signature. Composes cleanly when more end-label modes are added.

packages/openbridge-webcomponents/src/navigation-instruments/gauge-radial/gauge-radial.ts
anastasiyaBoyko: @talpitoo 
Done — added the comment above the arrow-property,
documents the binding intent so it doesn't get "fixed" into a method.

packages/openbridge-webcomponents/src/navigation-instruments/watch/watch.ts
anastasiyaBoyko: @talpitoo 
Thanks for the thorough flags 🙏 — addressed all of them here:
F1 — Added /** … Ignored when \zoomToFitArc` is true. /on all fourclipproperties inwatch.ts, and added the mutual-exclusivity note to the "Clipping for Partial Gauges" section of watch-radial-instruments.instructions.md`.
F2 — Added the usage hint to the clipLeft / clipRight JSDoc ("horizontal counterpart of clipTop/clipBottom, for quadrant/90° sectors"), and called out the obc-watch API extension in the PR description.
F3 — Went with the constant (the "best" option): extracted WATCH_DEFAULT_VIEWBOX = 448 in instrument-radial.ts, with a comment tying it to obc-watch's (176 + getPadding()) * 2 (default padding 48)
F4 — Added the ownership sentence to the arcFrame JSDoc ("if you pass arcFrame, you own keeping it in sync with areas / watchCircleType — obc-watch will not recompute it"). (Also corrected a follow-up precedence nit: arcFrame actually takes precedence whenever it's set — the if (this.arcFrame) branch runs before the zoomToFitArc one — so the wording reflects that.)
F5 — Added the INVARIANT comment above _bandRadius (additive everywhere; mixing additive/multiplicative offsets misaligns ticks, labels, advice masks and roundedArch arcs; proportional experiments removed).
F6 — Added a "Geometry Inputs Cheat-Sheet" section to the instructions doc: the property→effect table, the External / Internal / Max-min label model, and which combinations are validated vs undefined. Done here rather than as a follow-up.
F7 — Added a one-liner noting the textRadius → _bandRadius coupling is intentional (if a non-additive mode is ever reintroduced, inside labels follow it).

anastasiyaBoyko:
> The architectural shape of this PR is good — the additive-only `_bandRadius`, the viewBox-crop sector model, the design-aligned label modes, and the precomputed `arcFrame` handoff are all moves in the right direction. The remaining items are mostly about leaving enough breadcrumbs so the next person editing `watch.ts` doesn't accidentally break one of the invariants that this PR established.
> 
> For context: my previous review pass was primarily visual — confirming that sector layouts, needle, readout positioning, and tickmark behavior all looked correct across the variants. This round is mostly AI-assisted code-cleanliness review (JSDoc clarity, naming, API surface, future-proofing comments around the shared `obc-watch` / `obc-instrument-radial` machinery). Treat them as suggestions, not blockers — happy to land most of them as a follow-up if that keeps this PR focused.

@talpitoo 
Thanks — really appreciate both passes: the visual confirmation last round and this cleanliness/future-proofing sweep. The invariant-breadcrumbs point is well taken; that shared obc-watch / obc-instrument-radial surface is exactly where the next editor needs guardrails.
Rather than defer, I folded all of this round's suggestions into the PR — the rename, the width-derived inset, the JSDoc/::part clarifications, the clip* / arcFrame / _bandRadius invariant comments, the WATCH_DEFAULT_VIEWBOX constant, and a geometry-inputs cheat-sheet (with the External / Internal / Max-min model) in watch-radial-instruments.instructions.md. So the breadcrumbs live in-code now and no follow-up issue is needed. All of it is behavior-preserving — the only snapshot change is the new Sector180WideEndLabels regression story.
Thanks again for the thorough review 🙏

packages/openbridge-webcomponents/src/navigation-instruments/watch/watch.ts
talpitoo: Wow, even these, thank you :pray:!

talpitoo (approved):

ulrik-jo (approved):

PR #949, opened 2026-06-09
feat!: unify radial-instrument readouts

### LINKED ISSUE #945: 🟥 Maintenance/Follow-up: Radial instruments + Readout (PRs #903, #924, #936, #939, #942, #943)
# Follow-up: Radial instruments + Readout (PRs #903, #924, #936, #939, #942, #943)

> Investigation notes & proposed next steps after reviewing the six in-flight radial-instrument PRs. Visually the PRs look good — this issue focuses on **API consistency, geometry edge cases, and de-duplication opportunities** so that future maintainers don't get bitten by them.

## Why this issue

The six PRs listed below (#903, #924, #936, #939, #942, #943) all introduce the same feature in slightly different ways: embedding an `<obc-readout>` inside a watch-based radial instrument, often alongside new sector/scale variants and `zoomToFitArc` tweaks. Visually they look fine and each PR works in isolation, but reviewing them side-by-side surfaces a fair amount of **API drift, duplicated rendering code, and shared geometry edge cases** that we should resolve **before extending the same pattern to the remaining radial siblings** (`obc-rudder`, `obc-rate-of-turn`, and any future instrument that grows a centre readout). This issue is a consolidated review of those PRs with proposed next steps to bullet-proof and unify the API so the next round of work has one clear pattern to follow rather than five subtly different ones.

## TL;DR

All six PRs introduce variations of the same feature (`<obc-readout>` embedded inside a radial instrument) plus geometry / sector variants. They were merged to different branches piecemeal, so the resulting public API and rendering strategy is **inconsistent across siblings** and the implementations carry a fair amount of **duplicated logic** (readout HTML, scale fill/indicator color helpers, two-band advice generation, viewBox math, scale rotation tricks).

This is a tracking issue to:

1. Pick a single name and contract for the embedded readout property.
2. Extract repeated rendering helpers into shared modules.
3. Lock down `zoomToFitArc` edge cases (`arcAngle` clamping, clip normalization, container-pixel assumptions, stale `arcFrame`).
4. Make Storybook controls consistent (so QA and designers can toggle the same args across every radial instrument).
5. Decide whether/how `obc-rudder` and `obc-rate-of-turn` get the same readout treatment.

---

## 1. Per-PR summary

| PR | Title | Touched files | Net adds |
|----|-------|---------------|----------|
| [#903](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/903) | `obc-gauge-radial` sector-aware layout + readout | `watch.ts`, `tickmark.ts`, `instrument-radial.ts` (+css/stories), `gauge-radial.{ts,css,stories.ts}`, `readout.css` | `GaugeRadialSector` enum (`270`/`180`/`90-left`/`90-right`), `showReadout`/`label`/`unit`/`fractionDigits`, 4-edge clips on `obc-watch` and `obc-instrument-radial`, `endLabelsMaxMin`, `_bandRadius()` invariant, `defaultGaugeAngle` fallback, `normalizeClips` |
| [#924](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/924) | `obc-roll` readout + dual scale + priority | `roll.{ts,stories.ts}` | `ObcRollType.singleScale`/`dualScale`, `hasReadout`, `priority`, `WithReadout`/`DualScale`/`Enhanced` stories |
| [#936](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/936) | `obc-pitch` readout + dual scale + priority | `pitch.{ts,stories.ts}` | `ObcPitchType.singleScale`/`dualScale`, `hasReadout`, `priority`, same 3 stories |
| [#939](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/939) | `obc-pitch-roll` stacked readouts | `pitch-roll.{ts,stories.ts}` | `hasReadout`, stacked pitch+roll readouts, `WithReadout`/`ZoomedInWithReadout` stories |
| [#942](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/942) | `obc-rot-sector` readout | `rot-sector.{ts,stories.ts}` | `hasReadout`, `_readoutTopPercent` heuristic that lerps between 66–70% as `rotArcExtent` changes |
| [#943](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/943) | `obc-compass-sector` readout | `compass-sector.{ts,css,stories.ts}` | `hasReadout`, `_readoutTopPercent` from parsed `_cachedViewBox` (non-zoom) or fixed 70% (zoom) |

---

### 2.1 Readout boolean: `showReadout` vs `hasReadout`

| Component | Property | Source PR |
|-----------|----------|-----------|
| `obc-gauge-radial` | **`showReadout`** | #903 |
| `obc-pitch` | `hasReadout` | #936 |
| `obc-roll` | `hasReadout` | #924 |
| `obc-pitch-roll` | `hasReadout` | #939 |
| `obc-rot-sector` | `hasReadout` | #942 |
| `obc-compass-sector` | `hasReadout` | #943 |

Both are positive booleans (which is good — see [AGENTS.md § Boolean property naming](AGENTS.md#boolean-property-naming)) but **5-of-6 components agree on `hasReadout`**, parallel to `<obc-readout>`'s own `hasSetpoint` / `hasAdvice`.
PR #903 is the odd one out.

**Proposed:** rename `obc-gauge-radial.showReadout` → `hasReadout` before release. Repo-wide consistency check across siblings: `hasSetpoint`, `hasAdvice`, `hasReadout` everywhere.

### 2.2 Readout metadata API drift

Pitch / Roll / Pitch-Roll / ROT-sector / Compass-sector **hardcode** the readout's `label`, `unit`, `direction`, `variant`, `fractionDigits`,
`valuePriority`. Only `obc-gauge-radial` exposes `label`, `unit`, `fractionDigits` as instrument properties.

Consequence: consumers cannot localise or rename the readout label e.g. `Pitch` → `Tangage`, `ROT` → `Rateoftur`, `HDG` → `Cap`).

**Proposed:** every readout-bearing instrument should expose at minimum `label`, `unit`, `fractionDigits`, with the current strings as defaults.

### 2.3 Single-priority vs per-element-priority

| Component | API |
|-----------|-----|
| `obc-gauge-radial`, `obc-pitch`, `obc-roll`, `obc-rot-sector` | `priority: Priority` (single value) |
| `obc-pitch-roll` | `priority` + `priorityElements: PitchRollPriorityElement[]` |
| `obc-compass-sector` | `priority` + `priorityElements: CompassSectorPriorityElement[]` |

This is an existing inconsistency that the new readout PRs amplified (each readout uses a different rule for `valuePriority`: pitch-roll uses
`priorityFor(element)`, compass-sector uses `priorityFor(hdg)`, pitch/roll/rot use the flat `priority`, gauge-radial uses `priority` for value only).

**Proposed:** define a shared `PriorityElementBundle` mixin (or convention) that all multi-element instruments use; for single-element instruments `priority` stays as-is. Document the rule that `valuePriority` on the embedded readout mirrors the "primary" element.

### 2.4 Scale-variant enums

`ObcRollType` and `ObcPitchType` are two separate enums with identical members (`singleScale`, `dualScale`). Consider a single shared `ScaleVariant` enum exported from `navigation-instruments/types.ts`.

---

### 3.1 Embedded-readout block (5× near-identical)

Every PR (#924, #936, #939, #942, #943) ships essentially the same template:

```ts
${this.hasReadout
  ? html`<div class="readout">
      <obc-readout
        .variant=${ReadoutVariant.enhanced}
        .direction=${ReadoutDirection.vertical}
        .hasSetpoint=${false}
        .hasAdvice=${false}
        .value=${...}
        .fractionDigits=${0}
        .valuePriority=${this.priority}
        label="..."
        unit="..."
      ></obc-readout>
    </div>`
  : nothing}
```

…plus near-identical CSS (`position: absolute; left: 50%; transform:
translate(-50%, -50%);` and `::part(value-wrapper) { justify-self: center; }`).

**Proposed:** introduce one of:

- a small `renderInstrumentReadout({ label, unit, value, priority, topPercent })` helper in `navigation-instruments/readout/instrument-readout.ts`, or
- an `<obc-readout variant="instrument-center">` preset that handles positioning + part overrides internally, so individual instruments only
  pass data props.

The second is preferable because it puts the layout/parts coupling **inside** `obc-readout` — see § 5.

### 3.2 Pitch / Roll are mirror copies of each other

`pitch.ts` and `roll.ts` post-#936 / #924 are almost line-for-line the same file with `pitch`↔`roll`, `90°`↔`180°`, `cos`↔`sin` substitutions:

- `scaleFillColor` / `indicatorColor` getters — identical
- `renderScale(areas, opposite)` — identical body, only the rotation axis differs
- `private get advices()` — identical body, only `30°` (pitch) vs `45°` (roll) defaults differ
- center overlay SVG (horizon line + needle) — identical
- "opposite" `<obc-watch>` trick (`transform: rotate(180deg)` to render the same scale on the opposite arc) — identical
- styles — identical

**Proposed:** extract a `SingleAxisInclinometer` base / mixin (or a single `<obc-inclinometer axis="pitch|roll">` component) that takes:

- `arcCenter` (90 vs 180),
- `defaultAdviceOuter` (30 vs 45),
- `vesselImage` (side vs fore),
- transform composition for the indicator.

This would halve the surface for future bug fixes (any geometry fix today must be applied twice).

### 3.3 Pitch-Roll's 4× advice loop

`pitch-roll.ts` `get advices()` and `renderZoomedArcs()` already replicate the same caution-band logic four times (one per cardinal). Pull into a
single `cardinalAdvices(axis, center, max, trigger)` helper that yields the 4 bands.

### 3.4 ViewBox / clips logic

The default-sized viewBox formula `(176 + 48) * 2 = 448` now lives in:

- `watch.ts` — `(176 + this.getPadding()) * 2`
- `instrument-radial.ts` — `WATCH_DEFAULT_VIEWBOX = 448` (new in #903)

…and the `_radiusOffset = 0` / `_arcFrame = undefined` reset when leaving zoom mode is duplicated. The "ignore `clipTop/Bottom` when `zoomToFitArc`" rule is encoded both in `watch.ts` (silently ignored) and `instrument-radial.ts` (`this.zoomToFitArc ? 0 : clips.top`).

**Proposed:** export `WATCH_DEFAULT_VIEWBOX` (and `WATCH_DEFAULT_PADDING`)
from `watch.ts` and have all consumers import it. Document the "`clip*` ignored under zoom" rule in one place and remove the redundant
guards in consumers.

### 3.5 Two divergent codepaths in `advice.ts`

(Already flagged in `watch-radial-instruments.instructions.md`.) The "radiusOffset === 0" path and the "radiusOffset > 0" path render hatch
patterns differently to avoid snapshot churn. PR #903 didn't touch this but it's now hit by 4× more components (gauge-radial 90-sector etc.), so the risk of divergence has grown. Worth tracking as a separate cleanup later.

---

## 4. `zoomToFitArc` — edge-case checklist

Anything below is a place where the current implementation can either crash, visibly mis-render, or silently fall back to incorrect geometry.

### 4.1 `normalizeArcAngle` is inconsistently applied

`svghelpers/arc-frame.ts` clamps `arcAngle` to `[2, 180]` with a fallback (45 for pitch/roll/pitch-roll). But:

- `obc-rudder` doesn't pass `arcAngle` at all — it has a separate `maxAngle` prop. Confirm the rudder path doesn't get into a code branch that expects it.
- `obc-rot-sector` uses `rotArcExtent` (default 60) which is **not** passed through `normalizeArcAngle`; its `_readoutTopPercent` clamps to `[10, 60]` but the underlying scale doesn't.
- `obc-compass-sector` uses a fixed 120°; ignores `arcAngle`.

**Proposed:** make `normalizeArcAngle` the single entry point for **all** arc-half-extent inputs, with per-consumer fallback defaults documented in a table.

### 4.2 `MIN_ARC_HALF_DEG = 2` and `CORNER_GAP_PX = 32` in `pitch-roll.ts`

- `MIN_ARC_HALF_DEG = 2` silently floors degenerate inputs — fine but should be documented as an "edge cannot reach the indicator".
- `CORNER_GAP_PX = 32` is **in display pixels** but used inside the SVG coordinate system where 1 unit ≈ 1px **only at the default 400×400
  container**. At 200×200 the gap visually doubles; at 1000×1000 it shrinks to a sliver. Either:
  - rename and document as "SVG-unit gap", or
  - scale by `clientHeight / 400` at runtime.

### 4.3 `shiftArcFrameToOuterEdge` short-circuit

`shiftArcFrameToOuterEdge()` returns the input unchanged when the bbox is symmetric around the origin (`dist < 0.001`). This is correct for pitch-roll (4 axes balanced) but **silently disables outward-shift** for any future caller whose bbox happens to be centered. Add an assertion or a doc note where the helper is used.

### 4.4 Binary-search convergence in `computeZoomToFitArcFrame`

The 16-iteration doubling + 50-iteration bisection loop can fail to converge if the host container is smaller than `measureSize(0)` (i.e. the un-zoomed arc already overflows). The result is a returned `radiusOffset = 0` and a viewBox that's still too big. Currently no warning. Worth logging a single `console.warn` in dev mode when the search bottoms out without satisfying `measureSize(rOff) <= available`.

### 4.5 `arcFrame` ownership

PR #903 amended the `arcFrame` JSDoc to:

> If you pass `arcFrame`, you own keeping it in sync with `areas` / `watchCircleType` — obc-watch will NOT recompute it, so a stale frame renders stale geometry.

That's a real foot-gun: every consumer that computes its own frame (`pitch.ts`, `roll.ts`, `pitch-roll.ts`, `rudder.ts`, `instrument-radial.ts`)
must remember to invalidate on `areas`/`watchCircleType` change. Today they recompute on every `render()` which is correct but wasteful. **Proposed:** expose a stable hash of `{areas, watchCircleType, padding, clips, arcExtent}` and let `obc-watch` skip recomputation only when the hash matches.

### 4.6 Pitch / Roll `dual-scale` + `zoomToFitArc`

When both flags are true, two `<obc-watch>` siblings render — the second with `transform: rotate(180deg)` CSS. Only the first contributes its
`arcFrame` (they share `this._arcFrame`). The rotation is applied via CSS **after** SVG layout, so the rotated watch's viewBox is identical to the primary; visually that's correct. But: any zoom expansion that pushes the arc out of the *unrotated* bbox will also push the *rotated* arc out at the opposite side — make sure stories cover this combination at narrow arc angles. Currently `DualScale` stories don't set `zoomToFitArc: true`.

### 4.7 `obc-compass-sector` heading-dependent bbox

The `_readoutTopPercent` comment explicitly calls out that
`computeZoomToFitArcFrame` produces a `viewBox` whose `radiusOffset` /
position depends on the arc's *absolute* orientation (which cardinal axes it
crosses), even though the arc is always rotated back to the top for display.
The current fix is "use a fixed 70% under zoom". That's a workable patch
but a deeper fix would be:

- compute the frame in an axis-rotated bbox (the arc's local frame), or
- have the arc's bbox always treat the up axis as `(0, -1)`.

Worth opening as a separate ticket because it affects how readouts/labels
are anchored across compass-sector, rot-sector, and any future zoom-aware
component.

### 4.8 Readout-position heuristics

- `obc-rot-sector._readoutTopPercent` lerps between 66 % (wide) and 70 %
  (narrow) of host height as `rotArcExtent` slides 10°→60°. The endpoints
  were picked by eye, not derived from geometry.
- `obc-compass-sector._readoutTopPercent` parses the cached viewBox string
  with a `NaN` fallback.

Both should be replaced by a geometric anchor (centre of the inner ring's
visible chord) once the helper from §3.1 exists.

### 4.9 `gauge-radial` 90° sectors silently drop advices

```ts
.advices=${this.isSector90 ? [] : this.advices}
```

Consumers passing advices to a 90° gauge see nothing render and no warning.
Either log once in dev mode or refuse to render `90-left`/`90-right` with a
non-empty `advices`.

### 4.10 `normalizeClips` & opposite-clip collapse

`normalizeClips()` collapses an opposite pair (top+bottom or left+right)
whose sum exceeds 100 by zeroing **both**. That's safe but surprising: a
user who passes `clipTop: 60, clipBottom: 50` (intended one of them as a
typo) silently gets no clip at all. A `console.warn` in dev mode would help.

### 4.11 4-edge clips only honoured by `obc-watch` / `obc-instrument-radial`

`clipLeft` / `clipRight` were added in #903. The other instruments
(`obc-pitch`, `obc-roll`, `obc-pitch-roll`, etc.) still use `clipTop`/
`clipBottom` only — no functional regression but it's now an asymmetric API
across siblings.

---

## 5. `<obc-readout>` coupling risk

Every readout-bearing PR sets:

```css
obc-readout::part(value-wrapper) {
  justify-self: center;
}
```

and (compass-sector also) `::part(meta-wrapper)`. This reaches into
`obc-readout`'s internal DOM through Shadow Parts. If a future change to
`<obc-readout>` removes or renames those parts, **all five instruments
silently break**.

The user explicitly flagged this concern. Mitigations, in order of
preference:

1. **Add an "instrument-center" / "stack-centered" variant to
   `<obc-readout>`** that does the centering internally. Instruments stop
   touching parts.
2. If a variant doesn't fit, **freeze the part names with a CSS-only
   contract** — document `value-wrapper` and `meta-wrapper` as public API,
   add a snapshot story that uses both parts so a rename breaks tests.
3. Add a `defaultLabelOnly` mode that lets gauge-radial drop its current
   `.labelOnly` prop manipulation in `renderReadout`.

Until then, every PR that changes `<obc-readout>` DOM must be cross-checked
against the five consumers listed in this issue.

---

## 6. Storybook controls — consistency audit

The table shows which controls are explicitly declared in `argTypes` (✅),
auto-generated only (✳️ — present in `args` but no `argTypes` entry), or
missing entirely (❌).

| Property              | gauge-radial | pitch  | roll   | pitch-roll | rot-sector | compass-sector | rudder | rate-of-turn |
|-----------------------|--------------|--------|--------|------------|------------|----------------|--------|--------------|
| `hasReadout` / `showReadout` | ✅ `showReadout` | ✳️ | ✳️ | ✳️ | ✅ `boolean` | ✳️ | n/a | n/a |
| `priority`            | ✅           | ✅     | ✅     | ✅         | ✅         | ✅             | ✅     | ✅           |
| `zoomToFitArc`        | ❌ (n/a)     | ✳️     | ✳️     | ✳️         | ✅         | ✅             | ✅     | ❌ (n/a)     |
| `arcAngle`            | ❌ (n/a)     | ✳️     | ✳️     | ✳️         | n/a        | n/a            | n/a    | n/a          |
| `type` (scale variant)| n/a          | ✅     | ✅     | n/a        | n/a        | n/a            | n/a    | n/a          |
| `sector`              | ✅           | n/a    | n/a    | n/a        | n/a        | n/a            | n/a    | n/a          |
| `tickmarksInside`     | ✅           | n/a    | n/a    | n/a        | n/a        | n/a            | n/a    | n/a          |
| `label`/`unit`/`fractionDigits` | ✅ | ❌ (hardcoded) | ❌ | ❌ | ❌ | ❌ | n/a | n/a |

**Proposed clean-up before merge:**

- Add explicit `argTypes` entries for `hasReadout`, `zoomToFitArc`,
  `arcAngle` on pitch / roll / pitch-roll / compass-sector so QA can toggle
  them deterministically (instead of relying on autodocs inferring them).
- Once `label`/`unit`/`fractionDigits` are exposed on the per-instrument
  side (§ 2.2), surface them as `text`/`number` controls too.
- Add a `ZoomedInDualScale` story for both pitch and roll (currently
  uncovered).
- Add an `Enhanced` story to pitch-roll (currently only on pitch and roll).

---

## 7. Should `obc-rudder` and `obc-rate-of-turn` get the same treatment?

Both are watch-based, both can render the same value as text. Today neither
has `hasReadout`. Decision needed:

- **Yes:** apply the §3.1 helper / variant to both so the family is
  uniform.
- **No:** document explicitly *why* (e.g. rudder is always paired with a
  separate digital readout in real bridges).

Either way, the answer should be the same as for the rest of the family.

---

## 8. Suggested next steps (ordered)

1. **API freeze**
   - [ ] Rename `showReadout` → `hasReadout` on `obc-gauge-radial`.
   - [ ] Expose `label`, `unit`, `fractionDigits` on every readout-bearing instrument with sensible defaults.
   - [ ] Decide single-priority vs per-element-priority for the radial family; align names.
2. **Extract shared helpers**
   - [ ] Add a `renderInstrumentReadout()` helper or `<obc-readout variant="instrument-center">` preset (§ 3.1, § 5).
   - [ ] Extract `SingleAxisInclinometer` base for pitch/roll (§ 3.2) **or** at minimum factor out `cardinalAdvices()` and `renderOppositeScale()`.
   - [ ] Export `WATCH_DEFAULT_VIEWBOX` and remove the duplicated literal (§ 3.4).
3. **Bulletproof `zoomToFitArc`**
   - [ ] Funnel every `arc-half-extent` input through `normalizeArcAngle` (§ 4.1).
   - [ ] Convert `CORNER_GAP_PX` to either an SVG-unit constant or a client-size-aware computation (§ 4.2).
   - [ ] Log a dev-only warning when binary search bottoms out (§ 4.4) and when `normalizeClips` collapses an opposite pair (§ 4.10).
   - [ ] Replace readout-position heuristics with geometric anchors once the helper from §3.1 exists (§ 4.8).
   - [ ] Add a `90° sector + non-empty advices` warning on `obc-gauge-radial` (§ 4.9).
4. **Story consistency** (§ 6)
   - [ ] Add explicit `argTypes` for `hasReadout`, `zoomToFitArc`, `arcAngle` everywhere they apply.
   - [ ] Add `ZoomedInDualScale` and `Enhanced` stories where missing.
5. **Readout coupling**
   - [ ] Promote `obc-readout::part(value-wrapper)` and `::part(meta-wrapper)` to documented public API, OR add an
         `instrument-center` variant so consumers don't reach into parts (§ 5).
   - [ ] Add a snapshot story exercising both parts so future renames are caught at CI.
6. **Decisions**
   - [ ] Whether `obc-rudder` / `obc-rate-of-turn` get `hasReadout` (§ 7).
   - [ ] Whether `obc-compass-sector` heading-dependent bbox warrants the deeper fix in § 4.7.

---

## 9. Acceptance criteria for the follow-up work

- A single `hasReadout` boolean across the radial family, all defaulting to `false`.
- A single source of truth for the embedded readout markup (helper or variant) — `git grep "<obc-readout" packages/openbridge-webcomponents/src/navigation-instruments` should return one definition, not five.
- A single source of truth for `WATCH_DEFAULT_VIEWBOX` and the "`clip*` ignored under zoom" rule.
- `pitch.ts` and `roll.ts` share their renderer code (one file or one base class) — line count drops by ~40 %.
- `normalizeArcAngle()` is invoked by every consumer that takes an arc-half-extent input.
- Storybook: every radial instrument exposes the same nine toggles where applicable (`priority`, `state`, `hasReadout`, `zoomToFitArc`,
  `arcAngle`/`maxAngle`/`rotArcExtent`, `tickmarksInside`, `showLabels`, `label`, `unit`).
- No instrument reaches into `obc-readout`'s shadow parts without those parts being documented as public API (or, ideally, no reach-in at all).

---

## 10. References

- PR #903 — gauge-radial sector layout / readout (the foundational PR).
- PR #924 — roll readout / type / priority.
- PR #936 — pitch readout / dual-scale / priority.
- PR #939 — pitch-roll stacked readout.
- PR #942 — rot-sector readout.
- PR #943 — compass-sector readout.
- Instruction file:
  [.github/instructions/watch-radial-instruments.instructions.md](.github/instructions/watch-radial-instruments.instructions.md)
- Instruction file:
  [.github/instructions/setpoint.instructions.md](.github/instructions/setpoint.instructions.md)
- Core renderer: [packages/openbridge-webcomponents/src/navigation-instruments/watch/watch.ts](packages/openbridge-webcomponents/src/navigation-instruments/watch/watch.ts)
- Arc-frame helper: [packages/openbridge-webcomponents/src/svghelpers/arc-frame.ts](packages/openbridge-webcomponents/src/svghelpers/arc-frame.ts)
- Building block: [packages/openbridge-webcomponents/src/building-blocks/instrument-radial/instrument-radial.ts](packages/openbridge-webcomponents/src/building-blocks/instrument-radial/instrument-radial.ts)

talpitoo:
Introduce a shared `renderInstrumentReadout` helper and `?inline` CSS so all radial instruments (pitch, roll, pitch-roll, rot-sector, compass-sector, speed-gauge, gauge-radial) render their value readout consistently.

Extract a `SingleAxisInclinometer` abstract base shared by pitch and roll, reducing duplication; pitch and roll are now thin subclasses with SVG render hooks typed as `SVGTemplateResult`.

Expose `label`, `unit` and `fractionDigits` as public properties across the readout components and normalize their Storybook `argTypes` to match the gauge-radial reference set.

BREAKING CHANGE: the `showReadout` property is renamed to `hasReadout` on **gauge-radial** and **speed-gauge**. Update consumers and framework wrappers accordingly.

Closes #945

## Summary by CodeRabbit

* **New Features**
  * Configurable readout labels, units, and decimal places across navigation instruments.
  * Shared, consistent centered readout rendering and new single-axis inclinometer base with improved arc framing/zoom.

* **Breaking Changes**
  * Readout toggle renamed from showReadout → hasReadout.

* **Documentation**
  * Expanded Storybook controls for readouts, labels, units, decimals, and arc/zoom options.


packages/openbridge-webcomponents/src/navigation-instruments/readout/instrument-readout.ts
```diff
@@ -0,0 +1,91 @@
+import {html, type TemplateResult} from 'lit';
+import {classMap} from 'lit/directives/class-map.js';
+import './readout.js';
+import {
+  ReadoutDirection,
+  ReadoutStackVerticalAlignment,
+  ReadoutVariant,
+} from './readout.js';
```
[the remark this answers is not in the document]
talpitoo: done

packages/openbridge-webcomponents/src/navigation-instruments/readout/instrument-readout.css
```diff
@@ -0,0 +1,15 @@
+/* Shared obc-readout part overrides for readouts embedded inside radial
```
talpitoo, line 1: @tibnor i am going to leave a few walk-through comments for easier review.

This is a common CSS for the readout inside isntruments, similar to how we do it with the charts e.g. https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/blob/develop/packages/openbridge-webcomponents/src/bars-graphs/donut-chart/donut-chart.ts#L4

packages/openbridge-webcomponents/src/navigation-instruments/readout/instrument-readout.ts
```diff
@@ -0,0 +1,90 @@
...
+  labelOnly?: boolean;
+  centerValue?: boolean;
+  centerMeta?: boolean;
+  className?: string;
+  showZeroPadding?: boolean;
+  valueHasHintedZeros?: boolean;
+  minValueLength?: number;
+}
+
+/**
```
talpitoo, line 35: @tibnor this is a good overview of this file but basically readouts inside radial instruments share the same look&feel so this way the code remains DRY and central/single source of truth.

packages/openbridge-webcomponents/src/building-blocks/single-axis-inclinometer/single-axis-inclinometer.stories.ts
talpitoo, line 1: @tibnor this file removes the duplication from `pitch.ts` and `roll.ts` having it in a central place, DRY. It is similar to https://openbridge-next-storybook--949-feat-unify-radial-instr-i7fna9a6.web.app/?path=/docs/building-blocks-line-area-chart-base--docs where the 'Line Graph' and 'Area Graph' are implementing the abstract class.

talpitoo:
Ready for review!

ulrik-jo (approved):

PR #995, opened 2026-06-21
fix(automation-button): forward alert frame slots through abstract button

tibnor:
The abstract automation button rendered the underlying obc-automation-button
but only forwarded the badge slots. Slotted alert-frame content (icon, label,
timer) placed on a device such as obc-analog-valve never reached the alert
frame, so it could not be shown.

Forward the slots all the way down, remapping the public alert-frame-icon /
alert-frame-label / alert-frame-timer slots onto the automation button's
alert-icon / alert-label / alert-timer slots (which in turn feed the alert
frame's icon / label / timer slots). Document these slots on
ObcAbstractAutomationButton rather than on each device component.

Add an analog valve story that shows a bottom-flip alert frame with a custom
icon, an alert label and a clock, and add its visual baseline.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01XWMYJ6JF3jpspmmoQMNt6m

## Release Notes

* **New Features**
  * Enhanced alert-frame support by forwarding alert icon, label, and timer slot content for automation buttons, enabling fully customizable alert visuals.

* **Documentation**
  * Added new Storybook examples demonstrating alert-frame behavior with custom icons and bottom/side flip alert frame types.


talpitoo:
Before this leaves draft: the `@slot alert-icon` / `alert-label` / `alert-timer` tags sit on `ObcAbstractAutomationButton`, which has no `@customElement`. CEM emits one manifest entry per registered element and `lit labs gen` reads only the concrete class, so the slots work at runtime but produce no manifest entry and no React/Vue/Angular/Svelte binding ([`docs/agents/jsdoc.md`](../blob/develop/docs/agents/jsdoc.md#slots-and-events-are-consumer-critical-slot--fires)).

The concrete components (`analog-valve`, `digital-valve`, `converter`, `damper`, `fan`, …) carry no `@slot` tags at all, so the existing `badge-*` slots have the same gap. `lint:slots` passes because it only reads what each registered class's own file renders.

talpitoo:
#1034 also edits `abstract-automation-button.ts` — it adds `readoutValues` inline while this PR hoists the property docs into the class JSDoc. Whichever merges second needs a small conflict fix.

PR #1029, opened 2026-07-04
fix(automation-button): remove activated icon-holder background override

tibnor:
## Summary

The `activated` prop PR (#1026) added a CSS rule that changed the icon holder's border and background color in the activated state. This caused a visual regression — the icon holder should keep its normal background when activated.

This PR removes that override:

```css
.activated .icon-holder {
  border-color: var(--normal-activated-border-color);
  background: var(--normal-activated-background-color);
}
```

## Snapshots

The `switch-activated` and `valve-activated` visual baselines added in #1026 reflect the old (buggy) background. They need to be re-rendered against the corrected CSS via the `/update-snapshots` workflow.

---
_Generated by [Claude Code](https://claude.ai/code/session_015tZgr9X7QRphLokDGBu5D5)_

## Summary by CodeRabbit

* **Style**
  * Adjusted button styling by removing a specific activated-state icon appearance, resulting in a simpler visual state handling.

### DIFF
--- a/packages/openbridge-webcomponents/src/automation/automation-button/automation-button.css
+++ b/packages/openbridge-webcomponents/src/automation/automation-button/automation-button.css
@@ -247,11 +247,6 @@
   background: var(--automation-button-static-background-color);
 }
 
-.activated .icon-holder {
-  border-color: var(--normal-activated-border-color);
-  background: var(--normal-activated-background-color);
-}
-
 .alert-frame {
   position: absolute;
   top: 0;


tibnor:
/update-snapshots

---
_Generated by [Claude Code](https://claude.ai/code)_

tibnor:
/update-snapshots

---
_Generated by [Claude Code](https://claude.ai/code)_

tibnor:
/update-snapshots

---
_Generated by [Claude Code](https://claude.ai/code)_

tibnor:
Note for reviewers: the three `❌ Snapshot update failed` comments above are **not** caused by this change. I triggered `/update-snapshots` pre-emptively, but it failed each time on a pre-existing harness flake unrelated to this PR — every one of the 252 stories aborts at setup with `Failed to fetch dynamically imported module .storybook/vitest.setup.ts` (the vitest browser server not serving the setup module inside the Docker image), before any snapshot is compared.

It turns out **no snapshot update is needed**. The `Playwright Tests` visual job passes against the existing baselines: in the night theme captured by the linux baselines, `--normal-activated-background-color` (`rgb(55,55,55)`) is visually identical to the icon-holder's default appearance, so removing the `.activated .icon-holder` rule produces no pixel diff there. The activated-background regression was only visible in the light themes, which the snapshots don't capture. All required checks on this PR are green.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1034, opened 2026-07-06
feat(automation-button): add support for multiple readouts

tibnor:
## Summary by CodeRabbit

* **New Features**
  * Automation buttons can display multiple readout values with associated units.
  * Readouts support configurable decimal precision and digit counts.
  * Added an example showing multiple voltage and amperage readouts on a converter.

* **Improvements**
  * Readout values default to whole numbers when no decimal precision is specified.
  * Multiple readouts are presented together in a consistent stacked layout.


ludvigovrevik:
Heads-up: #1085 just changed the same `renderValue` in `automation-button-readout-stack.ts` that this PR touches, so a rebase will conflict there.

What changed on develop-bound #1085:
- The leading zero-padding now renders in a dimmed `<span class="pad">` (the design's ghost zeros), so there is no `paddedValue = zeroPadding + v` string anymore — the template emits `sign` + pad span + `digits` separately.
- Negative values render the minus sign **before** the padding, with the sign occupying one digit slot (`-5` at `nDigits: 3` → `-05`, `-30` → `-30`). This matches the convention #1097 establishes for `obc-readout` hinted zeros (`-12` at 4 digits → `-012`).

When composing `fractionDigits` on top of that, note the padding length is currently computed from the full `toFixed` string length — with fraction digits in play you'll likely want the pad to count only the integer part (mirroring #1097's `splitHintedValue`, where the decimal point and fraction never count toward the digit reserve).

talpitoo:
**`showReadoutOnOffState` is gone — use `showStatus`.** develop's `showStatus` on `ObcAbstractAutomationButton` gates the same early return, and `analog-valve`, `digital-valve` and the motorized button already guard on it, so keeping both would ship two properties for one behaviour. `converter.stories.ts` sets `showStatus: false` instead.

Say the word if you want it back as a separate control — it is one commit to restore.

`readoutValues` sits beside develop's `tag`, both gated on `showReadoutStack`.

talpitoo:
The readout stack now renders on `obc-readout-block` (#1097), so the baseline moved: padding zeros are hinted, not full strength. #1176 (finish the rebuild) and #1056 (decouple on/off state from the value readout) touch the same rows.

PR #1074, opened 2026-07-29
fix(charts): stop circular charts changing size on first mouse-over

### LINKED ISSUE #1061: <obc-donut-chart> : Donut chart increases size with the first mouse over
This concerns the "obc-donut-chart" component

**Bug description**

With the attribute "fixedHeight" the size of the component can be determined. This also determines if the labels are shown or not.
This also works correct. But with the first mouse over the donut changes size.

It seems it depends on the actual screen resolution how big the size change is and to which direction (bigger/smaller)

**Steps to Reproduce**

1. Open Storybook story "BARS and GRAPHS>Donut Chart>Half Donut"  
2. Set prop "fixedHeight" to approx. "300" (the labels just appear at this setting)
3. Hover the mouse over the donut
4. See error--> donut changes size

See small screen recording below:

Image

**Environment**

- Component version / branch: https://openbridge-next-storybook.web.app/
- Theme: all
- Framework: plain HTML
- Browser: Edge, Chrome and Firefox
- OS: Windows

 **Additional Context**

As Already mentioned it seems the actual screen resolution influences the size change. The screen recording is created on a 1920*1080 display. When using storybook on a much higher resolution screen the donut also changes size but becomes smaller and the size change is not so big.

**Side note**

The documentation states that when the property  "fixedHeight" >192 px the labels will appear. For the full donut this is correct but for the half donut the  "fixedHeight" value must be a lot higher

talpitoo:
Fixes #1061

## Root cause

The circular charts (donut, pie, polar, radial-bar) have a **fixed, self-computed size** — `fixedHeight` determines the circumference, the component computes `calculatedWidth × actualHeight` and applies it via `--chart-width`/`--chart-height` CSS variables. But the Chart.js instance still ran with `responsive: true, maintainAspectRatio: true, aspectRatio`, which fights that design in three compounding ways:

1. **Circular container measurement.** Chart.js's responsive mode measures the canvas's parent (`.wrapper`) — but the wrapper's size *derives from the canvas itself*, and with `legend` enabled it also includes the legend `<div>` sitting next to the canvas. Chart.js then re-derives the canvas height as `containerWidth / aspectRatio` and writes inline styles, growing the wrapper again. With the legend visible the canvas inflated to ~2× the intended size (e.g. computed layout `317×183` rendered as `627×363`), and the legend populating asynchronously (a rAF after chart creation) re-triggered this on every create.
2. **Deferred resize flushed on first hover.** `Chart.resize()` defers into `_resizeBeforeDraw` while the animator is running, and the animator can stop without a final draw (completed items are removed without drawing). The stale deferred resize then sits until the next `draw()` — which is triggered by the **first mouse-over**. Reproduced: `_resizeBeforeDraw` stayed pending for seconds after a `fixedHeight` change and was flushed exactly on hover.
3. **DPR-dependent rounding.** `getMaximumSize()`/`retinaScale()` round with `round1`/`Math.floor` while clamping height to the container, so the jump's direction and magnitude vary with `devicePixelRatio` — matching the reporter's observation that screen resolution changes the effect.

## Fix

- `responsive: false, maintainAspectRatio: false, devicePixelRatio: window.devicePixelRatio` for all four circular charts. The canvas render size is set explicitly from the computed layout in `createChart` (canvas width/height attributes) and applied via `chart.resize(w, h)` in `updateChart` (a no-op when unchanged). No container measurement, no aspect-ratio re-derivation, no deferred foreign resizes; DPR scaling stays crisp.
- **Issue side note fixed — half-donut label threshold.** A half donut's arc has the same diameter (`fixedHeight − 64`) as a full donut at the same `fixedHeight`; only the rendered height is halved. Labels were hidden until `fixedHeight ≈ 320` because the draw-time plugin gates re-derived the threshold from the rendered canvas height. The layout's `isTooSmall` stays based on `fixedHeight` (same 192px threshold for both modes, mimicking the full-donut calculation) and the donut's draw-time gates (outer labels, center readout, resize-observer threshold) now reuse that layout decision. Half-mode rendered height is also rounded to an integer.

## Regression coverage

- New shared helper `expectChartCanvasToMatchComputedLayout()` in `storybook-util.ts` asserts the canvas matches the component's computed layout (throws on unset CSS vars — CodeRabbit review fix).
- New `HalfDonutWithLegend` story reproduces the issue scenario and additionally asserts the size stays stable across the first mouse-over (failed with `canvas is 366x210.4 but computed layout is 334x192` before the fix).
- New `HalfDonutThresholdHeight` story locks the half-mode label threshold at `fixedHeight = 192`.
- The existing `WithLegend` stories of pie/polar/radial-bar now run the same layout assertion in their play functions.

## Deterministic visual tests (pre-existing flake root-caused)

`default-pie`/`default-polar` flaked between two render states ~4px apart because chart label widths are measured with `ctx.measureText()` at chart creation and never re-measured after the webfont loads — snapshot geometry depended on font-load timing. `vitest.setup.ts` now forces all registered font faces to load before each story renders; the affected baselines were refreshed to the loaded-font state. Four consecutive full-suite runs pass (37/37).

## Verification

- Playwright repros against Storybook at DPR 1 / 1.25 / 1.5 / 2, single slider change and drag sequences (including the recording's 207→167→303 sequence with legend): canvas now always equals the computed layout, no pending deferred resize, no size change on hover; tooltips still shown; Realtime story stable across data updates.
- All 37 storybook tests for the four charts pass repeatedly. Updated visual baselines: legend stories shrink to the intended geometry (previously captured the inflated canvas), full-circle charts gain 1–2 px as they now render at their exact computed size, and the flaky trio was refreshed to the deterministic loaded-font state. If CI's rendering disagrees, regenerate via `npm run update-snapshots:docker`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sizing consistency across donut, pie, polar, and radial bar charts by using fixed-height computed dimensions for canvas sizing and updates.
  * Prevented legends and resize behavior from causing canvas width/height drift.
  * Ensured full vs half-circle label/center rendering follows consistent fixed-height threshold logic.
* **Tests**
  * Added Storybook regression checks to confirm rendered canvas layout matches computed dimensions, including half-donut and legend scenarios.
* **Chores**
  * Improved Storybook determinism by waiting for document fonts to load before each story.


talpitoo:
Ready for review!

ulrik-jo (approved):

PR #1107, opened 2026-08-06
⛔ docs(readout): compare every instrument-field story against obc-readout

talpitoo:
:warning: DO NOT MERGE, for demo purposes only

> **Stacked on #1104** (`fix/1103-readout-digit-bounds`). That branch adds `readout-components.instructions.md`, which this page cites, so the base is set to it rather than `develop` — GitHub will retarget to `develop` automatically once #1104 merges. The diff below is this PR's single commit only.

## Why

`<obc-instrument-field>` has carried `@deprecated` since #663, with `<obc-readout>` as its successor — but nothing recorded **how much of it the readout actually replaces**. It has no consumers left in this repo, so the question only matters to external consumers planning a migration, and there was no way to answer it short of rebuilding each case by hand.

This PR does that rebuild once and commits it as a Storybook page.

## What

**`instrument-field.stories.ts`** — retitled to `Instruments/Instrument Field (deprecated)` and given a docs blurb pointing at the comparison. Stories and args are untouched. Baselines are keyed by *file path*, not story title, so all 16 existing snapshots still pass unchanged.

**`instrument-field-vs-readout.stories.ts`** (new) — one manual-mode story per legacy story: deprecated component on the left, closest `<obc-readout>` configuration on the right, verdict paragraph underneath. Args are deliberately not wired up; the point is a fixed, comparable pair.

Verdicts are `Fully covered` / `Differs` / `Not reachable`, where *Fully covered* means "no API gap", not "same pixels".

## Findings

**Not reachable (3)**

| Story | Gap |
|---|---|
| `HorizontalRegular` | At `horizontal` + `regular` the legacy component puts the tag **first** (tag │ setpoint │ value │ unit │ src). `obc-readout` always emits the meta zone after the value. `obc-readout-list-item` with `stacking="leading-unit"` covers the shape — rendered alongside for reference — but it is a different, row-oriented component. |
| `Off` | Legacy `off` dims value, label and source to `--element-inactive-color`. `off` never reaches the readout's root class list and no rule in `readout.css` / `readout-block.css` styles it, so the text stays at full value colour. The `off-value` slot also has no counterpart — `offText` takes a plain string. |
| `AlignMultiple` | `srcOptions.spaceReserver` and `unitOptions.spaceReserver` replace `--obc-instrument-field-source-width`, but there is **no label reserver**: `renderMetaZone()` calls `renderTextbox('label', …)` without one and no `labelOptions` exists. Ragged labels give each row a different width, so a right-aligned stack goes ragged. `obc-readout-list` only aligns `obc-readout-list-item`. |

**Differs (4)** — `EnhancedZeroPadding` (`maxDigits` counts total formatted characters in the legacy component but integer digits only in `obc-readout-block`; hint zeros also use different tokens), `NeutralColor` (legacy allows a grey value beside a blue setpoint; the readout ties both to `priority` by design), `Horizontal` / `HorizontalWithSrc` (no setpoint↔value divider), `VerticalCenter` (`center` centres only the label; `alignment="center"` centres everything), `AutoHideSetpoint` (`autoHideDeadband` is in value units; pop-up compares *rendered* strings, so a deadband is only expressible through `fractionDigits`).

**Shared restyle** — documented once on the docs page rather than repeated per story: cap-height vs `font-size` typography, SemiBold label, chevron setpoint marker, and — the most visible — `obc-readout` **always** demoting the setpoint one size tier, which the legacy component only does in one of its four layouts:

| Legacy layout | Legacy setpoint | Legacy value | Agrees? |
|---|---|---|---|
| vertical + `regular` | 16px | 16px | no |
| vertical + `enhanced` | 16px | 34px | **yes** |
| horizontal + `regular` | 16px | 16px | no |
| horizontal + `enhanced` | 34px | 34px | no |

The colour mapping, by contrast, is exact: `--element-neutral-enhanced-color` and `--instrument-enhanced-secondary-color` resolve to the same value in every theme, so the legacy default maps to `priority="enhanced"` and `neutralColor` to `priority="regular"`.

## Notes for review

- Every verdict was checked against its generated baseline PNG, not just reasoned about. Three notes were rewritten after the images contradicted them — most importantly `Primary`, which originally claimed only the setpoint marker differed.
- The 16 new baselines are the point of the page as much as the text: when a gap gets closed, the snapshot moves.
- No component code changed.

## Test plan

- [x] `npm run typecheck`
- [x] `npx eslint src/navigation-instruments/instrument-field/`
- [x] `npx vitest run --project storybook 'instrument-field'` — 32/32 pass (16 pre-existing + 16 new)
- [x] Baselines regenerated with the filter before `--update`; `git status` on `__vis__/` confirms zero collateral rewrites
- [x] Re-run without `--update` to confirm stability

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved readout handling for invalid, negative, fractional, non-finite, and excessively large digit settings.
  * Prevented excessive digit reservations and formatting operations from causing errors.
  * Standardized numeric width and zero-padding behavior across readout components.

* **Documentation**
  * Added guidance for readout component behavior and formatting rules.
  * Marked the legacy instrument field as deprecated and linked to readout alternatives.
  * Added comprehensive visual comparisons for migrating from instrument fields.

* **Tests**
  * Expanded coverage for digit limits, validation, formatting boundaries, and reserved-width behavior.


talpitoo:
Addressed in https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/1146

PR #1110, opened 2026-08-06
fix(wrappers): add missing event bindings for undocumented events

### LINKED ISSUE #1109: Undocumented events are missing from generated framework wrappers
## Problem

`@fires` tags are used by wrapper generators, not just documentation: `lit labs gen` reads them to emit the framework event props. A component that emits an event without a `@fires` tag gets **no `onX` binding** in the React/Vue/Angular/Svelte wrappers.

Verified: `obc-icon-button` has `@fires click` → its React wrapper has `onClick`. `obc-app-button`, `obc-command-button`, `obc-vendor-button` and `obc-notification-badge-button` lacked the tag → no `onClick` at all. Likewise `obc-poi-group`'s `obc-poi-group-target-released` was absent from its wrapper entirely.

## Why `lint:slots` didn't catch it

`npm run lint:slots` passes today, but has four blind spots:

1. **Generic type args.** `findDispatchedEvents()` in `script/check-slot-event-docs.ts` doesn't match `new CustomEvent<T>('name')`, so 9 dispatch sites are invisible. This hid `obc-poi-group-target-released`.
2. **Base classes.** Files without `@customElement` are skipped, hiding `obc-poi-data-layout-change` on `PoiBase`.
3. **Native events.** Never checked. 11 components expose only a native `click` from a passthrough `<button>` and documented none of it.
4. **Tag syntax.** 190 tags used `@fires name {Type} desc`. `cem analyze` only strips the type when it comes *first*, so 170 manifest descriptions began with a literal `{CustomEvent<…>}` and the payload type was lost from the `type` field. (`lit labs gen` parses both orders, so wrappers were unaffected.)

Separately, 45 tags had no description at all — blank cells in Storybook autodocs.

## Proposed fix

- Widen the `findDispatchedEvents()` regex to allow an optional generic type argument.
- Add the two missing `@fires` tags; add `@fires click` to the 11 native-click components.
- Normalize all tags to `@fires {Type} name - description`.
- Give every tag a description; target 0 warnings from `lint:slots`.

Note `obc-sequence-loading-spinner` must be excluded from the native-click set — its `<button>` is hardcoded `disabled`, so it can never fire a click.

## Known remaining artifact

`cem analyze` infers a phantom event literally named `eventName` for `obc-system-menu` and `obc-textarea-field`, from their dynamic `emit(eventName)` / `dispatchClickEvent(eventName)` helpers. It doesn't reach the wrappers, but it appears in `custom-elements.json`. Clearing it means changing runtime code to satisfy a doc tool — worth deciding separately.

ulrik-jo:
## Why

`@fires` tags are used by wrapper generators, not just documentation: `lit labs gen` derives the framework event props from them, so an undocumented event gets **no `onX` binding** in the React/Vue/Angular/Svelte wrappers.

`obc-icon-button` has `@fires click` → its wrapper has `onClick`. `obc-app-button`, `obc-command-button`, `obc-vendor-button` and `obc-notification-badge-button` did not → no `onClick` at all. `obc-poi-group`'s `obc-poi-group-target-released` was absent from its wrapper entirely.

## What

| Commit | Change |
| --- | --- |
| `fix(lint)` | The audit's dispatch regex rejected `new CustomEvent<T>('name')`, hiding 9 sites — one of them undocumented. Widened it, added the 2 missing tags, corrected a wrong claim in the module docblock. |
| `docs(events)` | `@fires click` on 12 components whose only public event is a native click from a passthrough `<button>`. |
| `docs(events)` | Normalized 190 tags to `@fires {Type} name - desc`. `cem analyze` only strips the type when it comes first, so 170 manifest descriptions carried a literal `{CustomEvent<…>}` and lost their payload type. |
| `docs(events)` | Descriptions for the 45 tags that had none. |

`obc-sequence-loading-spinner` is deliberately excluded from the native-click set — its `<button>` is hardcoded `disabled`, so it can never fire a click.

## Verification

- `npm run lint:slots` — 0 errors, 0 warnings (was 45 warnings)
- `custom-elements.json` — 0 type leaks
- `typecheck`, ESLint, Prettier — clean
- Generated the React wrappers to a temp dir: every newly tagged component now emits `onClick`, and `obc-poi-group` gains `onObcPoiGroupTargetReleased: … as EventName<CustomEvent<{target: Poi}>>`

Comment-only changes apart from one regex in `script/check-slot-event-docs.ts`; no visual snapshots affected.

## Known remaining

`cem analyze` infers a phantom event literally named `eventName` for `obc-system-menu` and `obc-textarea-field`, from their dynamic `emit(eventName)` helpers. It doesn't reach the wrappers. Clearing it means changing runtime code to satisfy a doc tool — left for a separate decision.

Closes #1109

## Summary by CodeRabbit

* **Documentation**
  * Standardized event documentation across web components with consistent event names, types, payloads, and descriptions.
  * Added documentation for previously undocumented click and custom events.
  * Clarified when interactive events are emitted, including disabled, static, and non-clickable states.
  * Documented event bubbling, composition, and payload details where applicable.
  * Improved event discovery for generated documentation, including support for generic event types.


packages/openbridge-webcomponents/script/check-slot-event-docs.ts
```diff
@@ -121,12 +129,14 @@ function hasDynamicSlot(code: string): boolean {
...
- * not mistaken for the component's own public API.
+ * not mistaken for the component's own public API. An explicit type argument
+ * (`new CustomEvent<{target: Poi}>('name')`) is optional and may itself contain
+ * one level of nested angle brackets.
  */
 function findDispatchedEvents(code: string): Set<string> {
   const events = new Set<string>();
   const re =
-    /this\.dispatchEvent\(\s*new\s+(?:Custom)?Event\s*\(\s*['"]([^'"]+)['"]/g;
+    /this\.dispatchEvent\(\s*new\s+(?:Custom)?Event\s*(?:<(?:[^<>]|<[^<>]*>)*>)?\s*\(\s*['"]([^'"]+)['"]/g;
```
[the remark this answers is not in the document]
ulrik-jo: Fixed — the diagnostic now reads ``dispatches a '<name>' event``.

packages/openbridge-webcomponents/src/automation/automation-tank/automation-tank.ts
```diff
@@ -122,6 +122,7 @@ export enum TankChartMode {
  * @slot alert-icon - Custom icon for the alert frame.
  * @slot alert-label - Label for the alert frame.
  * @slot alert-timer - Timer for the alert frame.
+ * @fires click - Fired when the tank is clicked. Not fired when `static` is set, which renders a non-interactive `<div role="img">` instead of a `<button>`.
```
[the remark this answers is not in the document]
ulrik-jo: Correct, and updated.

packages/openbridge-webcomponents/src/components/keyboard-numeric/keyboard-numeric.ts
```diff
@@ -87,9 +87,9 @@ const ALLOWED_CHARS_PATTERN = `^[${ALLOWED_CHARS.map((c) =>
  * `<obc-keyboard-numeric>` – An on-screen numeric keyboard with optional calculation and symbol modes.
  *
  * @slot leading-icon - Custom leading icon shown inside the input field (rendered when `hasLeadingIcon` is true).
- * @fires close-click {CustomEvent} Fired when the close button (or Escape) dismisses the keyboard.
- * @fires value-change {CustomEvent<{value: string}>} Fired whenever the current value changes.
- * @fires done-click {CustomEvent<{value: string}>} Fired when the DONE button (or Enter) is pressed.
+ * @fires {CustomEvent} close-click - Fired when the close button (or Escape) dismisses the keyboard.
```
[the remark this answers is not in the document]
ulrik-jo: Fixed — `close-click` dispatches with no `detail`, so the tag is now `{CustomEvent<void>}`.

packages/openbridge-webcomponents/src/components/number-input-field/number-input-field.ts
```diff
@@ -50,8 +50,8 @@ const baseFontSize = 16;
  * @slot leading-icon - Icon displayed before the input value (when `hasLeadingIcon` is true)
  * @slot label-icon - Icon displayed before the label text (when `hasLabelIcon` is true)
  * @slot helper-icon - Icon displayed before helper or error text (when `hasHelperIcon` is true)
- * @fires input {CustomEvent<{value: number}>} When the numeric value changes during editing
- * @fires change {CustomEvent<{value: number}>} When the value is committed on blur
+ * @fires {CustomEvent<{value: number}>} input - When the numeric value changes during editing
```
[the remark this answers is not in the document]
ulrik-jo: This appears to conflate two files.

The comment is anchored on `number-input-field.ts:53`, but describes `onNumberFieldInput` dispatching `{value: raw}`. That method is in `stepper-box.ts:245`, not this file.

Both are in fact consistent with their declared types:

- `number-input-field.ts` declares `{value: number}` and dispatches `detail: {value: this.value}`, where `value` is `@property({type: Number}) value = NaN` (line 59).
- `stepper-box.ts` declares `{value: string}` for its `input` event and dispatches the raw string read off the input element.

No change made. Happy to revisit if I've misread the finding.

packages/openbridge-webcomponents/src/components/slider/slider.ts
```diff
@@ -127,8 +127,8 @@ export type ObcSliderChangeEvent = CustomEvent<number>;
  * @slot icon-left - Slot for the left icon button (shown when `hasLeftIcon` is true)
  * @slot icon-right - Slot for the right icon button (shown when `hasRightIcon` is true)
  * @attr hugcontainer - If set, the slider will not have any spacing between the slider icons and the container
- * @fires value {ObcSliderValueEvent} - Fired when the value is changed
- * @fires change {ObcSliderChangeEvent} - Fired when user interaction completes and value has changed
+ * @fires {ObcSliderValueEvent} value - Fired when the value is changed
+ * @fires {ObcSliderChangeEvent} change - Fired when user interaction completes and value has changed
```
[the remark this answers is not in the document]
ulrik-jo: Fixed.

packages/openbridge-webcomponents/src/components/stepper-box/stepper-box.ts
```diff
@@ -59,10 +59,10 @@ export enum ObcStepperBoxType {
...
  * ```
  *
- * @fires down {CustomEvent<{value: number}>} Fired when the decrement (left or down) button is clicked
- * @fires up {CustomEvent<{value: number}>} Fired when the increment (right or up) button is clicked
- * @fires input {CustomEvent<{value: string}>} Fired when the user types in the number input field
- * @fires change {CustomEvent<{value: number | null}>} Fired when the numeric value changes from any source
+ * @fires {CustomEvent<{value: number}>} down - Fired when the decrement (left or down) button is clicked
+ * @fires {CustomEvent<{value: number}>} up - Fired when the increment (right or up) button is clicked
+ * @fires {CustomEvent<{value: string}>} input - Fired when the user types in the number input field
+ * @fires {CustomEvent<{value: number | null}>} change - Fired when the numeric value changes from any source
```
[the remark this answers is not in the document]
ulrik-jo: Fixed.

packages/openbridge-webcomponents/src/integration-systems/integration-dropdown-button/integration-dropdown-button.ts
```diff
@@ -184,8 +184,8 @@ export class ObcIntegrationDropdownButton extends LitElement {
   /**
    * Handles the dropdown-change and change event when a new option is selected. Updates the selected value and label, and dispatches a `dropdown-change` and 'change' event with the new selection.
    *
-   * @fires dropdown-change {ObcIntegrationDropdownButtonChangeEvent} - Fired when the user selects a different option.
-   * @fires change {ObcIntegrationDropdownButtonChangeEvent} - Fired when the user selects a different option.
+   * @fires {ObcIntegrationDropdownButtonChangeEvent} dropdown-change - Fired when the user selects a different option.
+   * @fires {ObcIntegrationDropdownButtonChangeEvent} change - Fired when the user selects a different option.
```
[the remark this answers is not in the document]
ulrik-jo: Fixed

jon-daeh (approved):

talpitoo:
@ulrik-jo a late AI review, in case you agree with some of it:

Two things must be fixed before merge (details inline): a mangled comment in `slide-button.ts` and a stray double-dash in `checkbox.ts`. Both are artefacts of the bulk rewrite.

Beyond that, three points about the shape of the change:

**1. The fix isn't enforced, and the repo docs still teach the broken form.**
`AGENTS.md:128-129` and `.cursor/rules/comments.mdc:200-201` both state "both tag orders are accepted", and the worked examples at `comments.mdc:63` and `comments.mdc:220` use the trailing-type form this PR just spent 190 edits removing. `parseEventTags()` / `findDocumentedEvents()` in the audit also still accept both orders. So nothing stops the next component from reintroducing the exact bug with CI green.

Suggest: make type-first the documented rule in both files, and have `lint:slots` warn (or error) on a trailing `{Type}`. Without that, this PR is a one-time cleanup rather than a fix.

**2. The audit has a hole that is the same class of bug this PR is about.**
`findDocumentedEvents()` matches `@(?:fires|event)` across the *entire file*. The analyzer does neither: it ignores `@event` completely, and only looks at the class-level block. So both of these pass `lint:slots` while producing no wrapper binding:
- an event documented only with `@event` at class level;
- an event documented only with `@fires` inside a method body.

Nothing in the tree hits this today (I checked — every `@event` here is inside a method body and each has a class-level `@fires` counterpart), but the audit is the stated safety net and currently can't catch the thing it's for. Restricting the scan to the class JSDoc block, and dropping `@event` from the accepted set, would close it.

**3. The description frames the normalization as a wrapper fix; it isn't.**
The analyzer accepts `@fires name {Type}` and `@fires {Type} name` equally, so the 190-tag normalization changes nothing in the generated wrappers. Its real payoff is the manifest (Storybook autodocs, IDE, playgrounds), which the PR body does note. Only `docs(events): document native click…` and the two added tags actually change wrapper output. Worth separating in the description so the intent is clear later.

**Minor:** the `@event` edits (18 of them, all inside method bodies) are invisible to every consumer, are inconsistent with each other — `system-button.ts`, `topbar-message-item.ts` and `toggle-button-vertical-group.ts` keep the summary line so the text is now duplicated, while `top-bar.ts`, `menu-button.ts` and `notification-message-item.ts` drop it — and `@event name - description` isn't valid `@event` syntax (the tag names an event, it doesn't take a `-` description). I'd revert that group rather than land it half-applied.

packages/openbridge-webcomponents/src/components/slide-button/slide-button.ts
```diff
@@ -190,8 +190,7 @@ export class ObcSlideButton extends LitElement {
   private completeSlide() {
     /**
      * Emitted when the slide action is completed (handle dragged past the
-     * threshold or activated via keyboard).
-     * @fires slide {ObcSlideButtonSlideEvent}
+     * @fires {ObcSlideButtonSlideEvent} slide - threshold or activated via keyboard).
```
talpitoo, line 193: @ulrik-jo The rewrite spliced this comment mid-sentence:

```js
/**
 * Emitted when the slide action is completed (handle dragged past the
 * @fires {ObcSlideButtonSlideEvent} slide - threshold or activated via keyboard).
 */
```

The original description ran across two lines and the tag was appended to the wrong half. Suggest:

```js
/**
 * @fires {ObcSlideButtonSlideEvent} slide - Emitted when the slide action is
 *   completed (handle dragged past the threshold or activated via keyboard).
 */
```

packages/openbridge-webcomponents/src/components/checkbox/checkbox.ts
```diff
@@ -96,8 +96,8 @@ export type ObcCheckboxChangeEvent = CustomEvent<{
  * ```
  *
  * @slot - No named slots.
- * @fires change {ObcCheckboxChangeEvent} – Emitted when the status changes.
- * @fires disabled {ObcCheckboxChangeEvent} – Emitted when the disabled state changes.
+ * @fires {ObcCheckboxChangeEvent} change - – Emitted when the status changes.
```
talpitoo, line 99: @ulrik-jo Both tags ended up with two separators — the inserted `-` plus the original en dash:

```js
* @fires {ObcCheckboxChangeEvent} change - – Emitted when the status changes.
* @fires {ObcCheckboxChangeEvent} disabled - – Emitted when the disabled state changes.
```

That leading `–` lands in the manifest description verbatim. Drop it from both lines. Worth grepping the branch for `- –` in case the rewrite hit any other en-dash descriptions.

packages/openbridge-webcomponents/src/components/navigation-item/navigation-item.ts
```diff
@@ -177,8 +177,7 @@ export class ObcNavigationItem extends LitElement {
   @query('obc-tree-navigation-item') private treeItemElement?: HTMLElement;
 
   /**
-   * Fired when the navigation item is clicked (either as a link or button).
-   * @fires click {CustomEvent<void>}
+   * @fires {CustomEvent<void>} click - Fired when the navigation item is clicked (either as a link or button).
```
talpitoo, line 180: @ulrik-jo This docblock now claims the component fires `click`, but the dispatch is on the global object, not the element:

```js
onClick() {
  dispatchEvent(new CustomEvent('click'));   // bare call → window.dispatchEvent
}
```

`onClick` is wired at lines 236 and 259, so every activation fires a synthetic `CustomEvent('click')` on `window` that no consumer of the element ever sees. What consumers actually receive is the native `MouseEvent` bubbling out of the inner `<a>` — so the tag's `{CustomEvent<void>}` type is wrong too.

Not introduced here, but the PR edits this exact block, and it's a good illustration of why the audit misses it: `findDispatchedEvents()` is anchored to `this.dispatchEvent(`, so a bare `dispatchEvent(` is invisible to it. Two suggestions: add `this.` (or delete the handler and rely on the native click), and consider having the audit flag bare `dispatchEvent(` inside a component class.

packages/openbridge-webcomponents/src/components/brilliance-menu/brilliance-menu.ts
```diff
@@ -212,8 +212,7 @@ export class ObcBrillianceMenu extends LitElement {
   }
 
   /**
-   * Handles palette selection changes and emits `palette-changed`.
-   * @fires palette-changed
+   * @fires palette-changed - Handles palette selection changes and emits `palette-changed`.
```
talpitoo, line 215: @ulrik-jo The method's summary line got folded into the tag, so the *event* description now describes the *handler*, and the method is left with no description:

```js
/**
 * @fires palette-changed - Handles palette selection changes and emits `palette-changed`.
 * @param {CustomEvent} event
 */
```

Same pattern at `stepper-box.ts:283` and `:303` ("Dispatches the `down` event when the decrement button is clicked") and `navigation-item-group.ts:149` ("Opens the flyout panel and emits the `open` event"). These read as method docs, not event docs.

Since the class-level tags already carry correct descriptions and these method-level tags are invisible to the analyzer, the cleanest fix is to restore the summary line and leave the bare `@fires name`:

```js
/**
 * Handles palette selection changes and emits `palette-changed`.
 * @fires palette-changed
 * @param {CustomEvent} event
 */
```

packages/openbridge-webcomponents/src/ar/poi-button/poi-button.ts
```diff
@@ -117,7 +117,7 @@ export interface ObcPoiButtonDataItem {
  *
  * ## Events
  * This component does not emit custom events.
- * Native `click` is available from the internal `<button>`.
+ * - Emits a standard `click` event (`onClick` handler in framework wrappers) when activated.
```
talpitoo, line 120: @ulrik-jo The prose section now contradicts itself and mixes a sentence with a bullet:

```
## Events
This component does not emit custom events.
- Emits a standard `click` event (`onClick` handler in framework wrappers) when activated.
```

The line this replaced ("Native `click` is available from the internal `<button>`") explained *why* there's a click without a custom event. Suggest:

```
## Events
This component emits no custom events. A standard `click` bubbles from the
internal `<button>` (`onClick` in the framework wrappers).
```

packages/openbridge-webcomponents/src/ar/poi/poi-base.ts
```diff
@@ -46,6 +46,9 @@ const VALID_POI_STATES = new Set(Object.values(ObcPoiState));
  * Implements `Poi` and contains all positioning, X-filter,
  * layout-change dispatch, and visual-query logic so that each variant
  * only needs to override `renderContent()` and `getVisualNodes()`.
+ *
```
talpitoo, line 49: @ulrik-jo What is `@ignore` doing here? `PoiBase` isn't a registered element, so it produces no manifest entry and no wrapper — and the three concrete subclasses (`poi-data`, `poi-vessel`, `poi-aton`) already carry their own `@fires obc-poi-data-layout-change`, which is what the "tags must live on the concrete class" rule requires.

If the tag is only here to satisfy the audit, that's worth saying in a comment, because as written it looks like duplicate documentation that a future reader might "clean up" from the subclasses instead.

packages/openbridge-webcomponents/src/components/app-button/app-button.ts
```diff
@@ -57,6 +60,8 @@ export enum AppButtonSize {
  * </obc-app-button>
  * ```
  *
+ * @fires click - Fired when the button is clicked (if not disabled).
```
talpitoo, line 63: @ulrik-jo Worth a smoke test in `react-demo` before merge. Declaring `click` changes how `onClick` is delivered in the React wrapper, not just whether it's typed.

In `@lit/react`'s `createComponent`, any prop in the `events` map is routed to `elementProps` and attached via `addEventListener` — it is *not* forwarded to `React.createElement`. Today `onClick` isn't in that map, so it falls through to React's synthetic event system (which does work on custom elements, just untyped). After this change it becomes a direct DOM listener.

Functionally equivalent for the common case, but propagation differs: `e.stopPropagation()` inside one of these handlers now runs at the element and stops the event before it reaches React's root delegation, which would silently kill `onClick` handlers on React ancestors. `obc-icon-button` has shipped with `@fires click` for a while, so the pattern is presumably fine in practice — but it's now on 12 more components at once. One click-through of the demo would settle it.

(applies to the whole native-click set)

packages/openbridge-webcomponents/src/components/slide-button/slide-button.ts
ulrik-jo: Confirmed, and it turned out to be one instance of a systematic bug rather than a one-off.

packages/openbridge-webcomponents/src/components/checkbox/checkbox.ts
ulrik-jo: Fixed. Grepped the branch for `- –` and `- —`: the two `checkbox` lines were the only hits, so no other en-dash descriptions were affected.

Root cause was the same normalizer assuming the description started right after the name, without checking for an existing separator.

packages/openbridge-webcomponents/src/components/navigation-item/navigation-item.ts
ulrik-jo: Confirmed — `dispatchEvent` unqualified in a class method resolves to `globalThis.dispatchEvent`, so every activation fired a `CustomEvent('click')` on `window` that no consumer of the element could observe.

Of your two suggestions I took the second (delete the handler), not the first. Adding `this.` would have made every activation fire **twice**: the native `click` already bubbles out of the inner `<a>` (and out of `obc-tree-navigation-item` in tree mode, since it is composed), which is what consumers have actually been receiving all along. So the handler and both `@click` bindings are removed, and the class-level tag drops its `{CustomEvent<void>}` type — what is delivered is the native event, matching the untyped `@fires click` on `obc-button` and `obc-icon-button`.

packages/openbridge-webcomponents/src/components/brilliance-menu/brilliance-menu.ts
ulrik-jo: Confirmed, and it is the same defect as the `slide-button` one

packages/openbridge-webcomponents/src/ar/poi-button/poi-button.ts
ulrik-jo: Adopted your wording

packages/openbridge-webcomponents/src/ar/poi/poi-base.ts
ulrik-jo: Fair point — as written it invites exactly the cleanup you describe. The tag is not needed by the audit either (the audit skips classes without `@customElement`, which is why the event was missing here in the first place); it was added for consistency with CLAUDE.md §3(c).

Rather than remove it I have made the intent explicit in the docblock:

```js
/**
 * The `@fires` tag below documents the dispatch site, which lives here rather
 * than in the variants. It is deliberately duplicated on each concrete
 * subclass — that is where `custom-elements.json` and the framework wrappers
 * pick the event up, since this class is not registered. Do not remove it from
 * the subclasses. `@ignore` keeps this class out of the generated docs.
 */
```

packages/openbridge-webcomponents/src/components/app-button/app-button.ts
ulrik-jo: Verified the mechanism in `@lit/react`'s source rather than by click-through, since the propagation question is decided there. `create-component.js`:

```js
if (eventProps.has(k) || k in elementClass.prototype) {
  elementProps[k] = v;   // → addOrUpdateEventListener, never reaches React.createElement
  continue;
}
```

So your description is exactly right: `onClick` moves from React's synthetic delegation to a direct DOM listener, handlers now receive a native `PointerEvent` instead of a `SyntheticEvent`, and `stopPropagation()` inside one will stop the event before React's root delegation and silently kill `onClick` on React ancestors.

Decision after discussing with @ulrik-jo: **keep it.** The Vue/Angular/Svelte bindings did not exist at all before, `obc-button` and `obc-icon-button` have shipped this way for a while, and the TypeScript type change surfaces any real breakage at compile time rather than at runtime. It will go in the release notes as a public-API semantics change for React consumers of the native-click set.

PR #1116, opened 2026-08-07
fix!: use web animation for blinking

tibnor:
BREAKING-CHANGE: the css custom properties `--alarm-blink-on` and similar is not animated any more. Use the `blinkingAll` function to bind the animation to the needed html element.

The motivation is that using a animation of inherited custom properties on the `html` node made the browser recalculating styles of all nodes every second. With this fix the custom variables are only animated on the needed components (alert-button, alert-icon and alert-frame), they are still syncronized.

## New Features
- Added blinking visual indicators for critical, alarm, warning, and low-priority alerts.
- Alert buttons, icons, and unacknowledged alert frames now animate according to alert severity.
- Added a Storybook example for a critical, unacknowledged alert frame.

## Bug Fixes
- Improved animation lifecycle handling so blinking stops cleanly when components are updated, disconnected, or reconnected.
- Prevented duplicate animations during repeated updates.
- Updated critical alert styling to use the appropriate blinking behavior.


packages/openbridge-webcomponents/src/palettes/variables.css
talpitoo, line 1: @tibnor this touches `variables.css`. Was this an automated Figma export or manual edit?

talpitoo (changes requested):

packages/openbridge-webcomponents/src/palettes/variables.css
tibnor: This was an manual fix. So the tool must be updated. Can you fix that?

talpitoo: @tibnor i haven't looked deeper into this issue. 
Are you suggesting that you want to have manual changes to `variables.css` and the tool should ignore/recognize those?

Or this manual change of yours should come from the tool itself automatically?

tibnor: I can fix it, that is fastest

ulrik-jo (approved):

talpitoo (approved):

PR #1136, opened 2026-08-13
feat(user-menu): make username and password optional fields

tibnor:
## Summary by CodeRabbit

* **New Features**
  * Added support for selecting the primary action in compact signed-in menus.
  * Added menu configurations for signed-in states without actions and sign-in states without recent users.
  * Recent-user sections now appear only when enabled and populated.
  * User profiles no longer display fallback recent-user data when profile information is unavailable.

* **Bug Fixes**
  * Improved spacing for sign-in-only menus.


talpitoo:
Spotted while merging `develop` in — not caused by the merge, and not something I changed.

The class JSDoc documents `showPassword` as applying to `user-sign-in` (`@availableWhen showPassword type in [signIn, userSignIn]`), but neither `user-sign-in` layout honours it:

- `renderUserSignInSmall` gates the **password** field on `this.showUsername`, so `showUsername: false` hides the password field and `showPassword: false` does nothing.
- `renderUserSignIn` (regular size) does not gate the password field at all.

The first one has a dead end behind it. `validateSignIn` computes `needsPassword` as `this.type === signIn ? this.showPassword : true`, so with `type="user-sign-in"`, `size="small"`, `showUsername: false` the password field is hidden while validation still requires a password — sign-in reports "Password is required" with no field to fill in.

Either gate both layouts on `showPassword` and fold `type === userSignIn` into `needsPassword`, or narrow the two `@availableWhen` lines to `type==signIn` if `user-sign-in` is deliberately out of scope.

Separately: `html` and `withActions` are still imported in `user-menu.stories.ts` after the custom `render()` and the decorator came out — two eslint warnings, the branch's only delta from the baseline.

talpitoo (approved):

jon-daeh (approved):

Issue #1138, opened 2026-08-13
Gauge Trend does not seem to resize to fit container


reach-mvbj:
After the the [fix ](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/1125)to the resizing issue for the Automation Tank we have run into an issue where the graphs inside the Automation Tanks have a lot of whitespace / padding around them.

<img width="1506" height="229" alt="Image" src="https://github.com/user-attachments/assets/6358bff9-24e1-42a0-ad61-078de4681279" />

Again we initially saw it in a wrapper around the OpenBridge package, but I have tried to make reproduction in pure JavaScript within some stories in Storybook:
https://openbridge-next-storybook--1137-bug-runaway-resizing-t-lunbtk1u.web.app/?path=/story/automation-tanks-tank--bug-chart-canvas-does-not-fill-cell
https://openbridge-next-storybook--1137-bug-runaway-resizing-t-lunbtk1u.web.app/?path=/story/instruments-gauge-trend--bug-css-zoom-applied-twice

From the reproduction it was suggested that it might be a problem in the underlying Gauge Trend that is used and might not be resizing when the container is resizes after intial render.

reach-mvbj:
It seems this bug could maybe be triggered by some side effects of some other CSS. However, we do not see it on any other instruments or other components? 

So it might be an unfortunate set of triggers from our project causing it, but it seems there still is a bug in the underlying component for the Gauge Trend (maybe others not used in our project).

PR #1140, opened 2026-08-14
⛔ feat(readout): add free content columns to the readout row and list

talpitoo:
:warning: do not merge, POC only

Adds `leading-content` / `trailing-content` slots to `obc-readout-list-item` and teaches `obc-readout-list` to keep those columns aligned across rows — so a readout row can carry an indicator on the left and a mini chart on the right, per the two Figma frames below.

This is both a **designer review** (six new stories) and a **POC answering one question: are slots sufficient?**

## The answer

**Slots are sufficient for the visuals, and insufficient on their own for the two things that actually matter.**

### 1. Column alignment — slots need a companion reserve

`align()` derives every existing reserver from *data* (`unit`, `maxDigits`, `fractionDigits`, `src`, `hasDegree`). Slotted content is light DOM the list cannot derive a width from, so a row with an indicator and a row without would staircase by 48px. Two rules fix that without ever measuring:

- **Width is a shared custom property** — `--obc-readout-list-item-leading-content-width` / `--obc-readout-list-item-trailing-content-width`, default 48px (the minimum touch target), identical on every row. Set it on the list to change every row and its spacers at once.
- **`align()` reads occupancy only** and sets `hasLeadingContentSpacer` / `hasTrailingContentSpacer` on the rows that leave a column empty while a sibling fills it. This is the `hasDegreeSpacer` trick, one column further out.

Measuring was deliberately rejected: `align()` writes properties that trigger a re-render, so a `getBoundingClientRect` inside it builds a measure → resize → measure loop — the same shape as the fixed-aspect chart sizing bug in #1138.

### 2. State sync — slots carry markup, not data

Nothing is forwarded into slotted content: not `value`, not `state`, not `priority`, not `dataQuality`. Arbitrary content has no known API to write to, so **a slotted indicator's value is a second binding the consumer owns** and must keep in step with the row's own `value`. Every story below binds the same number twice, on purpose, and the slot docs say so plainly — a reviewer who assumes the row drives the indicator would build a display that silently disagrees with itself.

If that trade turns out to be unacceptable, opt-in propagation (e.g. children marked `data-obc-bind="value state"` receive pushed state) is a purely additive follow-up. It is **not** in this PR.

## Why existing baselines did not move

The `<slot>` renders unconditionally; the spacer `<span>` renders only when reserved. An unfilled slot is `display: contents` and generates no box — so rows that slot nothing add no `.content` flex gap. Wrapping the slot in an always-present element instead would have added two gaps to every readout row in the library and moved the whole readout + automation-tank snapshot family. **Zero existing baselines changed.**

## Stories

**Instruments/Readout List**
- `WithIndicators` — pitch / roll / heave in the leading column (frame 116428). Values at both signs, since a compact indicator and its full-size twin only share a name.
- `WithTrendCharts` — `obc-graph-mini` sparklines in the trailing column (frame 116431).
- `WithTrendIndicators` — the same rows with `obc-gauge-trend-indicator` instead, so the open sparkline and the framed glyph can be compared side by side.
- `MixedSlots` — only some rows slot content; the reserve holds the labels and values on one line. Debug overlay on: solid purple = filled column, dashed = reserve.
- `TestSlotColumnReserve` — asserts the row that fills a column does **not** also reserve one, and that removing the last slotted element clears the reserve on the next pass.

**Instruments/Readout List Item**
- `SlottedContent` — leading / trailing / both / neither-but-reserved, and the three density tiers (the column width does not follow the size tier).
- `ColumnAlignmentWithContent` — without vs with reserves, the manual counterpart to what the list does automatically.

## Two questions for the designer

- **Frame 116428** stacks a small "State" line under the source and drops the label entirely. `src` is a single line today, so the state line is not reproduced. Should the source column become a two-line stack, the way `label`/`unit` already is under `leading-unit`?
- **Frame 116431** draws a grey baseline rule under each sparkline. `obc-graph-mini` has no such rule — it renders the line plus a trailing dot inside a bordered 32px box. Add the rule to `obc-graph-mini`, or does the framed box replace it?

## Not in this PR

`obc-readout` (the sibling layout) is untouched — this is layout-specific, not shared behaviour. The full `obc-gauge-trend` is deliberately not used in a row: it sizes itself from the box it measures, which is exactly the container shape that goes circular (#1138). Fixed-size indicators sidestep the question entirely.

## Verification

- `npm run typecheck`, `npm run lint` (0 errors; the 18 warnings are pre-existing and in unrelated files), `npm run lint:slots` (passed), `npm run analyze`, `npm run agents:sync`.
- `npx vitest run --project storybook 'readout-list'` — 31 passed. Existing baselines unmoved; the 7 new ones were reviewed as images before committing and held stable across four consecutive runs (the canvas-graph family is known to flake locally).
- `npx vitest run --project storybook automation-tank readout-block textbox` — 54 passed; `obc-automation-tank` slots readout lists, so it was checked too.

## Summary by CodeRabbit

* **New Features**
  * Added leading and trailing content columns to readout list items.
  * Preserved consistent row alignment with configurable fixed-width columns and automatic spacers for empty rows.
  * Supports indicators, charts, trend displays, and other custom content alongside readout values.
  * Added examples for mixed content, slot usage, and column alignment.

* **Documentation**
  * Clarified content-column alignment, spacer behavior, and consumer-managed slotted content.


talpitoo:
**Status check after the 6.1 merges: this POC is NOT superseded — kept open, rebased onto `develop` (`099bf316`).**

What the merged 6.1 work (#1145) did and did not absorb from here:

- **Not absorbed (the point of this PR):** the free `leading-content` / `trailing-content` columns, the shared custom-property width, and the `align()` occupancy spacers exist only on this branch — #1145 explicitly deferred the indicator/graph arrangements (Figma d3/d4 frames) to this POC. Nothing equivalent is on `develop`.
- **Absorbed — one of the two designer questions below is now answered:** *"Should the source column become a two-line stack (Source / State)?"* → yes; `srcOptions.deviation` (+ `srcOptions.state`) landed in #1145 as the two-line source block. The `WithIndicators` story here still binds a single-line `src`; when this PR moves past POC, its stories can adopt the merged two-line source instead of noting the gap.
- **Still open:** the `obc-graph-mini` baseline-rule question, and the state-sync trade-off (slots carry markup, not data — a slotted indicator's value stays a second consumer-owned binding).

Rebase notes: the component/list code applied cleanly on top of the 6.1 rework; the conflicts were only in the list-item stories/CSS where both branches appended sections (resolved keeping both — verified against this branch's original hunks byte-for-byte). Baselines regenerated on top of the merged 6.1 restyle (regular-weight labels etc.); the six content-column stories moved for exactly that restyle, `TestSlotColumnReserve`'s assertions still pass, and the run was stable across two consecutive verifies plus the `automation-tank` consumer (71/71).

PR #1195, opened 2026-09-04
fix: alert menu item long text wrapping

reach-mvbj:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved multi-line message items so descriptions wrap at word boundaries.
  - Long words in message descriptions can now break instead of forcing the layout to remain on one line.
  - Date information in multi-line items now aligns consistently with the item content.

- **Documentation**
  - Added alert menu item examples covering long text, unbroken words, shelved items, and secondary actions in narrow layouts.


packages/openbridge-webcomponents/src/components/message-menu-item/message-menu-item.css
```diff
@@ -137,7 +137,13 @@
...
     .text-container {
       grid-template-areas: "title date chevron" "description description description";
-      grid-template-columns: 1fr min-content min-content;
+      grid-template-columns: minmax(0, 1fr) min-content min-content;
+    }
+
+    /* Expanded items show the full title, so it has to wrap rather than widen the grid */
+    .title-container {
+      white-space: normal;
+      overflow-wrap: break-word;
```
talpitoo, line 145: Minor, but perhaps add the same rule to the description as well for consistency?

talpitoo (changes requested):
Minor visual tweaks:

Image

The time is cvertically centered but [in Figma](https://www.figma.com/design/kQMFf24Y1ry43HJWOStqd8/OpenBridge-6.1?node-id=9892-159325&t=PN7vtZgMsNSG1D6v-0) it keep its position top aligned - same line with the caret. Okay, Figma is not showing a use case for multiline title, still i think that we should keep the top alignment (time - caret - button).

packages/openbridge-webcomponents/src/components/message-menu-item/message-menu-item.css
reach-mvbj: Good catch, I have added that along with a story for it.

talpitoo (approved):
Looks good.

ulrik-jo (approved):

PR #1246, opened 2026-09-16
feat(user-menu, user-button): add role line and "Use another account" button

### LINKED ISSUE #1245: obc-user-button: aria-label from initials hides the visible label and role from screen readers
`obc-user-button` sets `aria-label` from the initials alone, so the button's
visible text is never announced:

```ts
aria-label=${this.initials || 'User button'}
```

A button rendering **Username** over **Captain** announces only "AB, button".
The `aria-label` overrides the element's text content, so neither `label` nor
the new `sublabel` reaches assistive technology.

### Why it matters

- **WCAG 2.1 AA § 2.5.3 Label in Name** — the accessible name must contain the
  visible label text. "AB" contains neither "Username" nor "Captain".
- **WCAG 2.1 A § 4.1.2 Name, Role, Value** — screen-reader users get initials
  where sighted users get a name and a role, so the recently-signed-in row in
  `obc-user-menu` announces as three buttons called "AB", "CD" and "EF" with no
  way to tell whose account is whose.

### Where

- `src/components/user-button/user-button.ts` — the `aria-label` binding
- Consumers affected: `obc-user-menu` (primary profile and the recent-user
  row), `obc-brilliance-menu`

### Suggested fix

Build the accessible name from the visible text when there is any, falling back
to the initials:

- name = `label` + `sublabel` when present (e.g. "Username, Captain")
- name = initials when the button has no label, which is the icon-only case the
  current fallback was written for

This changes the announced name for every existing consumer, which is why it is
filed rather than folded into the change that surfaced it. It needs a decision
on the exact phrasing — whether the role belongs in the name or in
`aria-describedby`.

### Scope note

The gap predates the `sublabel` property; adding a second visible line made it
more pronounced. Per `docs/agents/a11y.md`, a gap left in place on a touched
component is written down rather than silently inherited.

Not verified with a screen reader yet — § 8 item 5 is still outstanding for
this component.

ulrik-jo:
## Summary

`obc-user-button` gains an optional `sublabel`, a second line under the label. `obc-user-menu` uses it for the user's role and adds a "Use another account" button for user-sign-in layouts. 

## Why

Figma node `5261-45250` draws a Role line under every regular-size user button and a "Use another account" button under Sign in.

Closes #1245.

## What changed

- `obc-user-button`: `sublabel` property, `part="sublabel"`, hidden in static  mode like `label`.
- `obc-user-menu`: `userRole`, `role` on `ObcUserMenuUser`,  `showUseAnotherAccount` (default `true`, `attribute: false`), and  `use-another-account-click`. The menu stays controlled — switch `type` to  `sign-in` in the handler.
- Role renders on presence, no separate toggle, and only at regular size.
- Label weight: bold only when a sublabel sits under it. `selected` keeps its  bold label, where the weight marks selection.
- Large user button labels move from 12px to 16px, matching the Figma pairing.
- `aria-label` now carries the visible text instead of the initials (#1245).

- **`showRole` boolean** — dropped. The role text already answers the question;  a second switch would need keeping in sync.
- **Role rendered by the menu instead of the button** — the recent-user labels  live in the button's shadow DOM, so the menu would have to re-implement the  label stack.
- **Role at small size** — not drawn in Figma, so not invented.

## Verification

`npx vitest run --project storybook user-menu user-button brilliance-menu communication-table` — 47 pass, re-run without `--update` after regenerating.
`npm run lint` 0 errors. `npm run analyze` and `npm run wrappers` regenerated; `onUseAnotherAccountClick` is bound in all four wrappers.

Baselines moved: the no-role layouts lose their semibold, the large user buttons move to 16px, and both user-sign-in sizes gain the button.

a11y § 8: activation, focus ring and target size come from the native `obc-button`; accessible name checked in Chromium across label/sublabel/static/ initials-only. Not screen-reader tested.

## Docs

- [ ] docs/agents updated — not needed, no family rule changed
- [x] Comment pass done (`npm run lint:comments` clean for touched files)

## Follow-ups

- `use-another-account-click` has no test coverage; a snapshot cannot click.
- Long role text does not wrap or ellipsize, as with `label` today.

## Summary by CodeRabbit

- **New Features**
  - User buttons now support optional sublabels with accessible naming and disabled-state styling.
  - User menus can display user roles in profiles and recent-user entries.
  - Added an option to show or hide the “Use another account” action.
  - Small user-menu layouts omit role information.
  - Recent-user selection events now include the selected user’s role when available.
- **Documentation**
  - Added Storybook examples covering user-button sublabels, user roles, and “Use another account” visibility options.


packages/openbridge-webcomponents/src/components/user-menu/user-menu.ts
```diff
@@ -90,6 +96,9 @@ export type ObcUserMenuSignedInAction = {
  * ### Events
  * - `sign-in-click` – Fired when a sign-in button is clicked.
  * - `sign-out-click` – Fired when the sign-out button is clicked.
+ * - `use-another-account-click` – Fired when the "Use another account" button
+ *   is clicked. The menu keeps its own state; switch `type` to `sign-in` in
```
talpitoo: This reads as if the menu handles the switch itself. Suggest: "The menu does not change its own `type`; set it to `sign-in` in the handler to show the full form."

packages/openbridge-webcomponents/src/components/user-menu/user-menu.stories.ts
```diff
@@ -185,6 +190,57 @@ export const SignedInWithoutActions: Story = {
...
+/**
+ * Roles are set here but the small layout draws none, so this renders exactly
+ * as it would with no role at all.
+ */
+export const UserSignInSmallIgnoresRole: Story = {
+  args: {
+    type: ObcUserMenuType.userSignIn,
+    size: ObcUserMenuSize.small,
+    userRole: 'Role',
+    recentUsers: [{initials: 'AB', label: 'Username', role: 'Role'}],
```
talpitoo, line 237: `hasRecentlySignedIn` is off here, so this recent user never renders and its `role` is not exercised. Add `hasRecentlySignedIn: true` so the story shows both roles being ignored.

talpitoo (approved):
Looks good. Small change, matches the issue and the Figma note, and the aria-label fix is a good one.

Two nits inline, not blocking. One more, optional: `recent-user-click` still sends `{initials, label}` only. Passing `role` too would keep the event in sync with `ObcUserMenuUser`.

`static` mode still sets `aria-label` on a plain `<div>`, which screen readers skip. Not new here, fine to leave for the a11y round.

talpitoo (approved):

packages/openbridge-webcomponents/src/components/user-menu/user-menu.ts
ulrik-jo: Agree. Fixed in new commit

packages/openbridge-webcomponents/src/components/user-menu/user-menu.stories.ts
ulrik-jo: Fixed in new commit

jon-daeh (approved):

packages/openbridge-webcomponents/src/components/user-menu/user-menu.ts
```diff
@@ -133,11 +144,14 @@ export type ObcUserMenuSignedInAction = {
...
  * @availableWhen primaryActionId type==signedIn && size==small
+ * @property showUseAnotherAccount - Controls the visibility of the "Use another account" button.
+ * @availableWhen showUseAnotherAccount type==userSignIn
  * @slot signed-in-action-icon-<id> - Optional icon for a signed-in action, one per action; `<id>` is the normalized action id (shown in the `signed-in` type).
  * @fires {CustomEvent<{username?: string, password?: string}>} sign-in-click - Fired when a sign-in button is clicked.
  * @fires {CustomEvent<void>} sign-out-click - Fired when the sign-out button is clicked.
+ * @fires {CustomEvent<void>} use-another-account-click - Fired when the "Use another account" button is clicked.
  * @fires {CustomEvent<{id: string, label: string}>} signed-in-action-click - Fired when a signed-in action is clicked.
- * @fires {CustomEvent<{initials: string, label: string}>} recent-user-click - Fired when a recent user button is clicked.
+ * @fires {CustomEvent<{initials: string, label: string, role?: string}>} recent-user-click - Fired when a recent user button is clicked, carrying that user's entry.
```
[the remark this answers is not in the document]
ulrik-jo: Fixed

PR #1282, opened 2026-09-18
build(connector-diagram): make the package publishable to npm at 1.0.0

ulrik-jo:
## Summary

`packages/connector-diagram` could not be published: `npm pack` produced a
three-file tarball with no code in it. This adds the `prepack` build, the two
sibling licence files and the `repository`/`homepage` metadata, so a manual
`npm publish` ships a working package. No source change.

## Why

Eight deprecated components in `src/automation/**` — `corner-line`,
`horizontal-line`, `vertical-line`, `three-way-line`, `end-point-line`,
`line-cross`, `direction-line`, `line-overlap` — carry `@deprecated` notes
reading "Please use the `@oicl/connector-diagram` package instead", and the
migration guide shipped with them (#1262). `npm view @oicl/connector-diagram`
returns 404: the package has never been published, so the guidance points
nowhere.

Publishing it as it stood would have been worse than not publishing. `dist/` is
gitignored and no lifecycle script built it, so the tarball was `README.md`,
`LICENSE.txt` and `package.json` — 2.6 kB, no code, and `main`/`types`/`exports`
all pointing at files that were not there.

This closes the last of PR #1243's Follow-ups that is actionable without picking
a release model.

## What changed

`packages/connector-diagram/package.json` only:

- **`prepack: "npm run build"`** — `npm publish` and `npm pack` compile `dist/`
  first. The core solves the same problem the same way.
- **`files`** gains `LICENSE.txt`, `LICENSE-AGPL.txt`, `LICENSE-APACHE.txt`. npm
  auto-includes `LICENSE.txt` alone, and that file is the one that refers to the
  other two — the dual AGPL/Apache-with-change-date grant is unreadable without
  them. The core lists all three explicitly for the same reason.
- **`repository`** (with `directory: packages/connector-diagram`) and
  **`homepage`**, matching the core. `repository` is also the prerequisite for
  npm provenance if this package ever moves into a workflow.

The version goes to **`1.0.0`**: the name is claimed permanently on first
publish, and `0.x` would tell consumers the API may still break while the
deprecation notes send them here.

`publishConfig.access: public` was already present.

`docs/agents/connector-diagram.md` gains a **Publishing** section — the manual
route, why `prepack` exists, why the licence files are listed, and that a local
publish gets no provenance — with its three generated adapters re-synced.

- **Adding a `pkgRoot` entry to `.releaserc.json`.** The least wiring, but it
  locks the version to the core's: `0.1.0` would jump to `2.0.0-next.N` and the
  package would re-publish on every core release even when untouched. Its
  release cadence should follow its own code. `ci-and-release.md`'s statement
  that the package is not among the six released packages stays true.
- **A dedicated `release-connector-diagram.yml` workflow.** The right end state,
  and it would restore provenance, but it is a larger change and the deprecation
  notes are live now. The metadata here is a prerequisite for it either way, so
  nothing in this PR is thrown away by doing it later.
- **Publishing `0.1.0` as it stood.** The version the package was added with,
  but a `0.x` line advertises an unstable API to exactly the users the
  `@deprecated` notes are redirecting, and the first published version is not
  retractable. `1.0.0` states the intent the deprecation notes already imply.

## Verification

```
typecheck   npm run typecheck (packages/connector-diagram): 0 errors
unit        npm test: 6 files, 41 tests passed
pack        rm -rf dist && npm pack --dry-run
            before: 3 files, 2.6 kB — README.md, LICENSE.txt, package.json
            after:  49 files, 43.0 kB — dist/ rebuilt by prepack, README,
                    all three licence files
consumer    packed the tarball and npm install'd it into a clean scratch project
            outside the workspace:
              import {renderSegments, themeFromCss, DEFAULT_THEME}
              → function, function, present
              tsc --strict on the README's usage example, including
              `import type {Segment}`: 0 errors
agent docs  npm run lint:agents: up to date
format      prettier --check on the four changed Markdown files: clean
```

The pre-existing "over 300 lines" warnings from `lint:agents` are for six other
docs; `connector-diagram.md` is not among them.

A root `npm install` rewrites four wrapper entries in `package-lock.json` from
`2.0.0-next.158`/AGPL-3.0-only back to `2.0.0-next.154`/Apache-2.0, because the
generated wrapper manifests on disk lag the last release — the behaviour
`connector-diagram.md` § Dependencies already documents. Unrelated to this
change; the lockfile is not in the diff.

## Docs

- [x] docs/agents updated — `connector-diagram.md` § Publishing, adapters synced
- [x] Comment pass done — no source files in the diff; `lint:comments` does not
      cover this package

## Follow-ups

Deliberately not done here, and intentionally not filed as issues:

- No publish workflow, so the first publish is manual and gets no npm
  provenance, unlike the six packages `release.yml` handles.
- `@oicl/openbridge-webcomponents` stays `"*"` and `overrides` stays a no-op in a
  workspace member, both from #1243's Follow-ups.
- The package-level `package-lock.json` still records the pre-workspace specs and
  is unused by the workspace install.

## Summary by CodeRabbit

- **New Release**
  - Connector Diagram is now available as version 1.0.0.
  - Package distribution includes the required license information and built output for reliable installation.
  - The package is configured for public npm distribution.

- **Documentation**
  - Added guidance for publishing the package, verifying release contents, managing independent versions, and understanding local publication limitations.
  - Documented the build and packaging checks performed before publication.


packages/connector-diagram/package.json
```diff
@@ -1,8 +1,14 @@
 {
   "name": "@oicl/connector-diagram",
-  "version": "0.1.0",
+  "version": "1.0.0",
```
tibnor, line 3: Jeg trodde jeg foreslo 0.1.0, slik at den har en experimental status?

tibnor (approved):

Issue #1106, opened 2026-08-06
⬜ Maintenance: Map Storybook tags e.g. `WIP` with JSDoc `@experimental`, `@beta`, etc.


talpitoo:
# Storybook lifecycle tags from JSDoc

## 1. Problem

A component's lifecycle state is recorded in two places that do not agree.

**In code**, [PR #1024](https://github.com/Ocean-Industries-Concept-Lab/openbridge-webcomponents/pull/1024) (`9d1c9148`, 2026-07-01) put a class-level JSDoc tag on every documented `@customElement`:

| Tag | Class-level count | Meaning (from that commit) |
| --- | --- | --- |
| `@stable` | 125 | production-ready, stable API |
| `@beta` | 57 | feature-complete, API may still change |
| `@experimental` | 51 | early stage, API likely to change |
| `@deprecated` | 11 (42 incl. member-level) | slated for removal |

**In Storybook**, `meta.tags` plus the badge definitions in `.storybook/manager.ts` tell a different, much sparser story. Badges exist for only three tags — `5.0`, `alpha`, `wip` — and `5.0` is used by zero story files.

Resolving each story's `meta.component` to the class that registers it (258 of 264 story files have one) gives the exact drift:

| Code says | Storybook says | Count |
| --- | --- | --- |
| `@stable` | *(nothing)* | 125 |
| `@beta` | *(nothing)* | 56 |
| `@experimental` | *(nothing)* | 37 |
| `@experimental` | `wip` | 12 |
| `@experimental` | `alpha` | 9 |
| `@deprecated` | *(nothing)* | 10 |
| `@beta` | `wip` | 1 |
| `@beta` | `alpha` | 1 |
| *(no tag in code)* | *(nothing)* | 6 |

Storybook badges 21 of 51 experimental components, badges none of the 10 deprecated ones, and uses `wip` and `alpha` interchangeably for states the code distinguishes. Nothing prevents further drift.

### Constraint discovered during investigation

`cem analyze` natively lifts `@deprecated` into `custom-elements.json` (51 occurrences) but **silently drops `@stable`, `@beta` and `@experimental`** — they appear zero times in the manifest. Any consumer of those three must parse the TypeScript source itself or add a CEM plugin. This rules out a manifest-driven solution without extra machinery.

### Latent bugs found

[...]

PR #1283, opened 2026-09-18
chore(lint): stop reporting lit-localize's idle disable directives

talpitoo:
## Summary

Clears the four `Unused eslint-disable directive` warnings in the generated locale files, without deleting the directives. `npm run lint:eslint` goes from 17 warnings to 13 on `develop`; the remaining 13 are the lifecycle tags and the story parameter that #1260 handles.

Two lines of config. Opened as a draft because the trade-off below is a team call, not because anything is unfinished.

## Why

`lit localize build` writes both directives into every locale file from a fixed template in `@lit/localize-tools/lib/modes/runtime.js`:

```js
/* eslint-disable no-irregular-whitespace */
/* eslint-disable @typescript-eslint/no-explicit-any */
```

It emits them unconditionally, for every locale, regardless of content. ESLint 9's flat config defaults `reportUnusedDisableDirectives` to `warn`, so when neither rule finds anything to suppress the directives are reported as unused — 2 directives × 2 locale files = 4 warnings.

The directives are not wrong, they are pre-emptive, and both cases are plausible for translated output:

- `no-irregular-whitespace` — translations legitimately carry non-breaking spaces, narrow no-break spaces and ideographic spaces. Spanish and Finnish happen not to right now.
- `@typescript-eslint/no-explicit-any` — lit-localize's runtime output can carry `any` in the `str` / `html` plumbing. These two files happen not to.

So the right outcome is "stop reporting them while idle", not "remove them".

These four are also the reason `npm run lint` has never read as clean, which is how the count drifted from 4 to 342 unnoticed before #1260.

## What changed

One `linterOptions` line added to the block that already scopes a rule to these files:

```js
{
  // Generated locale files use string-literal unions from lit-localize
  files: ['**/generated/locales/*.ts'],

  // lit-localize writes both disable directives into every locale file
  // whether or not the translations need them, so an idle one is not a
  // stale directive to delete — a translation carrying a non-breaking
  // space would put it back to work.
  linterOptions: {reportUnusedDisableDirectives: 'off'},

  rules: {
    'openbridge/prefer-enum-over-string-literal-union': 'off',
  },
},
```

- **Delete the directives from the generated files.** They come straight back on the next `lit localize build`, and `generated-code.md` forbids hand-editing generated output.
- **Ignore `src/generated/**` wholesale**, the way `eslint.comments.config.mjs` does. Heavier than needed: it would stop the locale files being linted at all. The `prefer-enum-over-string-literal-union` entry directly above shows this file already prefers a targeted off-switch for exactly these paths.
- **Patch lit-localize upstream** so the template emits a directive only when the content needs one. Correct in principle, but a third-party round trip for a cosmetic warning.
- **Leave the warnings.** Defensible — they are harmless — but they keep `npm run lint` permanently noisy, which is the condition that let the warning count grow unnoticed.

## What the team should weigh

Turning `reportUnusedDisableDirectives` off for these files means a directive that becomes genuinely stale — say lit-localize stops emitting `any`, or the rule is removed from the config — will no longer be flagged there. That is the whole cost, and it is bounded to two generated files nobody edits by hand.

If you would rather keep the nag, close this and the four warnings stay as the accepted baseline.

## Verification

- `npm run lint:eslint` — 17 → 13 on `develop`; zero `Unused eslint-disable directive` warnings remain.
- **The directives still suppress.** Injecting a `U+00A0` into a translated string in `es-419.ts` produces no `no-irregular-whitespace` report, confirming the first directive is still doing its job rather than being bypassed. Reverted after the check; no generated file is modified by this PR.
- `npm run lint` exit 0, `npm run typecheck` clean, `npm run format:check` clean, `npm run test:rules` 103 passed.
- No source, manifest or baseline changes — the diff is `eslint.config.mjs` only.

## Docs

- [x] docs/agents updated — not needed; no rule or family behaviour changed, and the lint command set is unchanged.
- [x] Comment pass done

## Follow-ups

- With this and the four `obc-indicator-*` lifecycle tags (#1106 deferred them to the automation family owner, being decided now and landing in #1260), `npm run lint` reports **zero** warnings.


tibnor:
I think this is an good idea. Until we put focus into translations

talpitoo:
@tibnor thanks for confirming, then it is ready for review!

tibnor (approved):

